PR #2, opened 2024-01-10
Design for centralized provider version management

asilverman:
## Overview

REP based on [#anthony-c-martin  proposal](https://github.com/Azure/bicep/issues/12202) to use `bicepconfig.json` to simplify provider declaration syntax.

### DIFF
--- a/active/0002-centalized-mgmt-resource-types-provider-versions.md
+++ b/active/0002-centalized-mgmt-resource-types-provider-versions.md
@@ -1,6 +1,6 @@
 ---
 REP Number: 0002
-Author:asilverman (Ariel Silverman)
+Author: asilverman (Ariel Silverman)
 Start Date: 2023-12-19
 Feature Status: Private Preview
 Bicep Issue Number(s): [12202](https://github.com/Azure/bicep/issues/12202)
--- a/active/0002-centalized-mgmt-resource-types-provider-versions.md
+++ b/active/0002-centalized-mgmt-resource-types-provider-versions.md
@@ -10,7 +10,7 @@ Bicep Issue Number(s): [12202](https://github.com/Azure/bicep/issues/12202)
 
 ## Summary
 
-This document proposes a feature that enables users to manage Bicep resource type provider versions from a central location. This enhances convenience by sparing the user from making changes in multiple files in order to upgrade a dynamically loaded resource types provider package version.
+This document proposes a feature that enables users to manage Bicep resource type provider versions from a central location. This improves convenience by sparing users from making changes in multiple files in order to upgrade a dynamically loaded resource types provider package version.
 
 ## Terms and definitions
 
@@ -25,7 +25,7 @@ The [resource type providers dynamic loading feature](https://github.com/Azure/b
 
 ```bicep
 provider 'br:mcr.microsoft.com/bicep/providers/az@1.2.3' // fully-qualified
-provider 'br/public:az@1.2.3' // uses provider aliasing defined in bicepconfig.json
+provider 'br/public:az@1.2.3' // with provider aliasing defined in bicepconfig.json
 ```
 
 We also currently support the following syntax for "built-in" providers:
@@ -48,15 +48,15 @@ The current syntax (shown above) presents several challenges:
 1. The new syntax is quite verbose, and cannot be controlled centrally. For users with a large set of files, picking up the latest version of `az` will require modifying every single file.
 1. The aliasing (assuming that the last segment of the registry namespace == the provider name) & versioning (having an `@` character instead of a `:` which is familiar to module users) behavior of the new syntax can be confusing.
 1. Its deceiving and confusing to allow `sys` to have a pinned version fixed to `1.0.0` - this **should** imply that there are no changes between Bicep versions, which isn't realistic.
+1. It's an "either/or" - either users follow the current behavior (`az` imported by default) **or** they're using the new syntax. We don't have a plan for how to combine the two.
+1. With [Radius](https://github.com/radius-project/radius), it's conceivable that we may want users to **not** have `az` imported by default. We want to allow users to opt-out from the implicit importing of the `az` namespace 
 
-- It's an "either/or" - either users follow the current behavior (`az` imported by default) **or** they're using the new syntax. We don't have a plan for how to combine the two.
-- With [Radius](https://github.com/radius-project/radius), it's conceivable that we may want users to **not** have `az` imported by default.
-
-### Proposed Changes
-
-- [Resolution of namespace identifiers by VSCode extension when inspecting published module sources](#resolution-of-namespace-identifiers-by-vscode-extension-when-inspecting-published-module-sources)
-- [Wildcard semantics for provider versions in `bicepconfig.json`](#wildcard-semantics-for-provider-versions-in-bicepconfigjson)
-
+## Proposed Changes
+<!-- no toc -->
+- [Client side changes](#client-side-changes)
+  - [Add a new syntax for the provider import declaration statements: `provider {providerName} [as {optionalAlias}]`](#add-a-new-syntax-for-the-provider-import-declaration-statements-provider-providername-as-optionalalias)
+  - [Change the delimiter '@' used by the supported provider declaration syntax to ':' to align with module reference syntax and increase consistency](#change-the-delimiter--used-by-the-supported-provider-declaration-syntax-to--to-align-with-module-reference-syntax-and-increase-consistency)
+  - [Allow users to configure providers that are implicitly imported](#allow-users-to-configure-providers-that-are-implicitly-imported)
 ## Detailed design
 
 ### Client side changes
@@ -70,38 +70,82 @@ provider az
 provider kubernetes as k8s // uses optional aliasing of the
 ```
 
-The resolution of the symbols will be resolved from the builtin `bicepconfig.json` under a new section:
+The resolution of the identifiers will be determined from inspection of the closest `bicepconfig.json` to the file merged with the builtin configuration. We propose to add a new section `providers` and `implicitProviders`.
 
+Example: The new default configuration
 ```json
 {
   // prior bicepconfig.json sections
   // ...
   "providers": {
-    // If you specify the entry `az` here you override the builtin `az` provider
     "az": {
-      // A package reference string (below) gestures Bicep to source the types from the registry address.
+      "builtIn": true
+    },
+    "kubernetes":{
+      "builtIn": true
+    },
+    "microsoftGraph":{
+      "builtIn": true
+    }
+  },
+  "implicitProviders": [
+    "az"
+  ]
+}
+```
+
+Users can opt-in into the centralized provider version management by specifying a `bicepconfig.json` in their project structure.
+
+Example: A `bicepconfig.json` that defines a dynamically loaded provider
+```json
+{
+ "providers": {
+    "az": {
+      "source": "mcr.microsoft.com/bicep/providers/az",
+      "version": "0.2.3"
+    },
+  },
+  "implicitProviders": [
+    "kubernetes"
+  ]
+}
+```
+
+Bicep will replace-merge the contents in the `bicepconfig.json` with the default resulting the th following configuration being loaded to be used in the compilation.
+
+`bicepconfig.json`
+```json
+{
+ "providers": {
+    "az": {
       "source": "mcr.microsoft.com/bicep/providers/az",
       "version": "0.2.3"
     },
     "kubernetes": {
-      // Using the value "builtin" (below) gestures Bicep to source types from the statically compiled NuGet package reference
+      "builIn": true
+    },
+    "microsoftGraph":
+    {
       "builtIn": true
     }
-  }
-  // ...
+  },
+  "implicitProviders": [
+    "kubernetes"
+  ]
 }
 ```
 
-Since legacy provider declaration syntax grammar continues to be supported, we introduce the following constraints:
+Given legacy provider declaration syntax grammar continues to be supported, we introduce the following constraints:
 
 - To ensure consistency with pre-existing handling of `bicepconfig.json` the configuration file closest to the Bicep file in the directory hierarchy is used.
+- The `az` identifier is implicitly imported into the global scope, the author can override the type definitions used in the namespace by adding an entry in the `providers` section (as shown above). 
+- The builtIn versions used for the resolution of built-in providers are determined by the NuGet package reference dependency in `Bicep.Core.csproj`.
+- The keys of the `providers` object must be distinct. Notice how the source is the full repository path, so its possible to disambiguate in the case an author chosses to consume a provider with the same name from separate sources.
+- The keys of the `providers` section will be the identifiers (`{providerName}`) in the `provider {providerName} [as {optionalAlias}]` declaration statement.
 - The `sys` identifier is coupled to the Bicep bits and its version cannot be overriden or dynamically uploaded using the mechanism described above
-  If a user specifies an entry called `sys` in the section above, this will be identified by the json schema as forbidden value. The reason is to prevent someone from overriding this reserved identifier.
-- The `az` identifier is implicitly imported into the global scope, the author can override the type definitions used in the namespace by adding an entry in the `providers` section (as shown above).
-- The keys of the `providers` object must all be unique. Notice how the source is the full repository path, so its possible to disambiguate in the case an author chosses to consume a provider with the same name from separate sources.
-- The keys of the `providers` section will be used as the identifiers in the `provider {providerName} [as {optionalAlias}]` statement
+- If a user specifies an entry called `sys` in the `providers` section above, it will result in a json schema violation. This behavior is to prevent a user from overriding the sys namespace.
 
-The syntax below currently used to import built-in providers will be deprecated and result in a diagnostic as a consequence of the proposed changes in this proposal
+Using the syntax below to import built-in providers will be deprecated and result in a diagnostic error as a result of the implementation of this proposal
 
 ```bicep
 provider 'sys@1.0.0'
@@ -117,10 +161,14 @@ provider 'br:mcr.microsoft.com/bicep/providers/az:0.2.3' // notice its NOT using
 provider 'br/public:az:0.2.3' as myAz
 ```
 
+#### Allow users to configure providers that are implicitly imported
+
+The `implicitProviders` section of `bicepconfig.json` allows users to opt-in/out of implicit import functionality for a given provider. This section also follows the existing config merge semantics so users interested in keeping with the legacy behavior must opt-in by specifying the `az` provider in the array.
+
 ## Drawbacks
 
-- Costumers may be confused by the configt merge semantics behavior
-- Costumers will not be able to tell from inspecting sources alone what is the provider version used to resolve resource types
+- Users may be confused by the configt merge semantics behavior
+- Users will not be able to tell from inspecting sources alone what is the provider version used to resolve resource types
 
 ## Alternatives
 
@@ -197,7 +245,7 @@ This is especially problematic for 3rd party namespace identifiers specified in 
 
 ### Resolution of namespace identifiers by VSCode extension when inspecting published module sources
 
-The ability to inspect sources is being implemented in Bicep to allow users to see what sources were used to compile a Bicep module, since the version information is codified in the `bicepconfig.json` and that is not published with the sources, the costumer will not be able to immediately know what version of a provider was used to resolve the resources in the souces file. This is illustrated in the example below.
+The ability to inspect sources is being implemented in Bicep to allow users to see what sources were used to compile a Bicep module, since the version information is codified in the `bicepconfig.json` and that is not published with the sources, the user will not be able to immediately know what version of a provider was used to resolve the resources in the souces file. This is illustrated in the example below.
 
 **Example: User consumes a published module (that uses new provider syntax) and inspects its sources**
 
@@ -251,8 +299,4 @@ At this time we are not proposing support for wildcard semantics on provider ver
 
 ### Deciding if `sys` is a compile time import or a resource types provider package
 
-It is possible to think of `sys` more like a compile-time namespace import rather than a provider, since all it contains are some built-in functions. We used to share the keyword import between providers and imports, but now that we changed the keyword for providers, at some point we will need to update the implementation for sys to make it a compile-time import. This is out of scope of this REP however.
-
-### Implicit import of provider identifiers to global scope
-
-In the current impementation two provider identifiers are implicitly imported: `sys` and `az`, their types are available in the global scope and the author doesn't neeed to dereference types using the identifier (i.e. `resourceGroup() vs az.resourceGroup()`). A prior version of this proposal introduced the concept of `importProviderByDefault` to allow users to appy the aforementioned behavior on other providers. This is out of scope for this proposal and is not advised since pollution of the global namespace makes the code harder to read.  
+There is an ongoing discussion on the topic of `sys` being a compile-time import or a resource types provider. Until the topic is settled, we will treat it as a provider and allow users the option to override the identifier `sys` if they chose in `bicepconfig.json`. Doing so will effectively result in users losing access to basic bicep functions, although this scenario is unlikely considering current usage patterns.
--- a/active/0002-centalized-mgmt-resource-types-provider-versions.md
+++ b/active/0002-centalized-mgmt-resource-types-provider-versions.md
@@ -62,7 +62,7 @@ The current syntax (shown above) presents several challenges:
 
 ### Client side changes
 
-#### Add a new syntax for the provider import declaration statements: `import {providerName} [as {optionalAlias}]`
+#### Add a new syntax for the provider import declaration statements: `provider {providerName} [as {optionalAlias}]`
 
 The example below shows a usage of the proposed syntax
 
--- a/active/0002-centalized-mgmt-resource-types-provider-versions.md
+++ b/active/0002-centalized-mgmt-resource-types-provider-versions.md
@@ -54,9 +54,9 @@ The current syntax (shown above) presents several challenges:
 
 ### Proposed Changes
 
-[Add a new syntax for the provider import declaration statements: `import {providerName} [as {optionalAlias}]`](#add-a-new-syntax-for-the-provider-import-declaration-statements-import-providername-as-optionalalias)
-[Remove support for syntax `import '{providerName}@{providerVersion}' [as {optionalAlias}]`](#remove-support-for-syntax-import-providernameproviderversion-as-optionalalias)
-[Change the delimiter '@' used by the supported provider declaration syntax to ':' to align with module reference syntax and increase consistency](#change-the-delimiter--used-by-the-supported-provider-declaration-syntax-to--to-align-with-module-reference-syntax-and-increase-consistency)
+* [Add a new syntax for the provider import declaration statements: `import {providerName} [as {optionalAlias}]`](#add-a-new-syntax-for-the-provider-import-declaration-statements-import-providername-as-optionalalias)
+* [Remove support for syntax `import '{providerName}@{providerVersion}' [as {optionalAlias}]`](#remove-support-for-syntax-import-providernameproviderversion-as-optionalalias)
+* [Change the delimiter '@' used by the supported provider declaration syntax to ':' to align with module reference syntax and increase consistency](#change-the-delimiter--used-by-the-supported-provider-declaration-syntax-to--to-align-with-module-reference-syntax-and-increase-consistency)
 
 ## Detailed design
 
--- a/active/0002-centalized-mgmt-resource-types-provider-versions.md
+++ b/active/0002-centalized-mgmt-resource-types-provider-versions.md
@@ -6,7 +6,7 @@ Feature Status: Private Preview
 Bicep Issue Number(s): [12202](https://github.com/Azure/bicep/issues/12202)
 ---
 
-# Title - Centralized management of resource type provider versions using package.json
+# Title - Centralized management of resource type provider versions using bicepconfig.json
 
 ## Summary
 
@@ -54,7 +54,6 @@ The current syntax (shown above) presents several challenges:
 
 ### Proposed Changes
 
-- [Updating `bicepconfig.json` merge semantics](#updating-bicepconfigjson-merge-semantics)
 - [Resolution of namespace identifiers by VSCode extension when inspecting published module sources](#resolution-of-namespace-identifiers-by-vscode-extension-when-inspecting-published-module-sources)
 - [Wildcard semantics for provider versions in `bicepconfig.json`](#wildcard-semantics-for-provider-versions-in-bicepconfigjson)
 
@@ -103,7 +102,7 @@ Since previous provider declaration syntax forms continue to be supported, we in
 - To ensure consistency with pre-existing handling of `bicepconfig.json` the configuration file closest to the Bicep file in the directory hierarchy is used.
 - The `sys` namespace is coupled to the Bicep bits version and cannot be overriden or dynamically uploaded using the mechanism described above
 
-#### Remove support for syntax `import '{providerName}@{providerVersion}' [as {optionalAlias}]`
+If a user specifies an entry called `sys` in the section above, this will be identified by the json schema as forbidden value. The reason is to prevent someone from overriding this reserved identifier.
 
 The syntax below currently in use to import built-in providers will become invalid and result in a diagnostic as a result of this change
 
@@ -123,7 +122,7 @@ provider 'br/public:az:0.2.3' as myAz
 
 ## Drawbacks
 
-- Costumers may be confused by the config merge semantics behavior
+- Costumers may be confused by the configt merge semantics behavior
 - Costumers will not be able to tell from inspecting sources alone what is the provider version used to resolve resource types
 
 ## Alternatives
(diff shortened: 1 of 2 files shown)


active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,247 @@
...
+- **types.json protocol** - It is a serialization protocol used by Bicep to encode type information as data that can be loaded to the Bicep runtime.
+
+## Motivation
+
+The [resource type providers dynamic loading feature](https://github.com/Azure/bicep/issues/5453) introduces a change in provider declaration statement syntax to include the registry address and version that Bicep will use to resolve resource type declarations.
+
+** Provider Declaration Syntax**
+
+```bicep
+import 'br:mcr.microsoft.com/bicep/providers/az@1.2.3' // fully-qualified
```
anthony-c-martin, line 27: [nit]The syntax has since changed to `provider`

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,247 @@
...
+
+- It's an "either/or" - either users follow the current behavior (`az` imported by default) **or** they're using the new syntax. We don't have a plan for how to combine the two.
+- With [Radius](https://github.com/radius-project/radius), it's conceivable that we may want users to **not** have `az` imported by default.
+
+### Proposed Changes
+
+[Add a new syntax for the provider import declaration statements: `import {providerName} [as {optionalAlias}]`](#add-a-new-syntax-for-the-provider-import-declaration-statements-import-providername-as-optionalalias)
+[Remove support for syntax `import '{providerName}@{providerVersion}' [as {optionalAlias}]`](#remove-support-for-syntax-import-providernameproviderversion-as-optionalalias)
+[Change the delimiter '@' used by the supported provider declaration syntax to ':' to align with module reference syntax and increase consistency](#change-the-delimiter--used-by-the-supported-provider-declaration-syntax-to--to-align-with-module-reference-syntax-and-increase-consistency)
```
anthony-c-martin, line 59: [nit] formatting is broken
```suggestion
* [Add a new syntax for the provider import declaration statements: `import {providerName} [as {optionalAlias}]`](#add-a-new-syntax-for-the-provider-import-declaration-statements-import-providername-as-optionalalias)
* [Remove support for syntax `import '{providerName}@{providerVersion}' [as {optionalAlias}]`](#remove-support-for-syntax-import-providernameproviderversion-as-optionalalias)
* [Change the delimiter '@' used by the supported provider declaration syntax to ':' to align with module reference syntax and increase consistency](#change-the-delimiter--used-by-the-supported-provider-declaration-syntax-to--to-align-with-module-reference-syntax-and-increase-consistency)
```

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,247 @@
...
+```bicep
+import 'br:mcr.microsoft.com/bicep/providers/az:0.2.3' // notice its NOT using az@0.2.3
+import 'br/public:az:0.2.3' as myAz
+```
+
+## Drawbacks
+
+> Discuss any potential drawbacks associated with this approach. Why might this _not_ be a suitable choice? Consider implementation costs, integration with existing and planned features, and the migration cost for Bicep users. Clearly identify if it constitutes a breaking change.
+
+1. The proposal uses `bicepconfig.json` to declare the provider information and is ruled by `bicepconfig.json` merge semantics. The merge semantics for `bicepconfig.json` specify that for a given Bicep source file, the configuration file closest to the Bicep file in the directory hierarchy is used. This behavior is not intuitive and may cause users that have specified multiple config files in their project file-structure to experience unexpected behaviors that may include build errors when Bicep attempts to resolve a namespace identifier. This is illustrated by the example below.
```
jeskew, line 123: I would consider this a separate discussion. We could adopt the proposed changes whether or not `bicepconfig.json` allows merges, and there's no reason this proposal should be gated on that work.

asilverman: agreed, I listed this item in the out of scope, but I did want to call out the drawback that results of this merge semantics not existing

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,247 @@
...
+
+```bicep
+{
+     "experimentalFeaturesEnabled": {
+        "compileTimeImports": false
+     },
+}
+```
+
+The user may intuitively expect Bicep to resolve the provider details of `modFooProvider` by inspection of `src/bicepconfig.json` and for this project to compile successfully in a way that feels familiar with other programming languages where merge semantics are applied on code dependencies (e.g. NPMs `package.json`), however, Bicep config merge semantics uses the configuration file closest to the Bicep file in the directory hierarchy and merges them to the builtin config.
```
anthony-c-martin, line 177: W.r.t familiarity - is this true? I didn't know that NPM had merge semantics.

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Alternatives
+
+- Using a separate file other than `bicepconfig.json` to store resource type provider version information was evaluated but adds complexity to UX as users need to be aware of two separate merge semantics (one for config and one for provider packages) 
+
+## Rollout plan
+
+This change affects the client side only and is applied only when the pre-existing feature flag `DynamicTypeLoadingEnabled` is set. 
+
+## Unresolved questions
```
alex-frankel, line 249: do we need to discuss wildcard semantics in this proposal somewhere? It doesn't necessarily need to be in the first iteration

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,247 @@
...
+
+> - What aspects of the design do you expect to resolve through the REP process before merging?
+> - What parts of the design do you anticipate resolving through feature implementation before stabilization?
+
```
jeskew, line 253: One open question for me here is, what about version matchers and lock files? That seems like a question for a future REP but one we should be mindful of lest any one-way doors be taken in implementing this REP.

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: I need to double check, I may be mistaken here for sue as I am not a JavaScript expert. In any case the example is to illustrate and idea. Perhaps I can find a better example for this

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,247 @@
...
+In this case Bicep will try to resolve the `modFooProvider` details from `src/modules/bicepconfig.json`. The config doesn't specify a `providers` section, and the builtin config doesn't include a definition for provider `foo` and so the compiler cannot resolve the provider declaration statement and will emit a diagnostic.
+
+If we replace `foo` with `az` in the example above, Bicep resolves `modFooProvider`'s version to be the version of `az` specified in the builtin configuration and not the version `1.2.3` specified in `src/modules/mod.bicep` (inherited from `src/bicepconfig.json`) which is also not intuitive. The reson for this behavior is that current `bicepconfig.json` merge semantics merge `src/modules/bicepconfig.json` with the builtin config and the version applied would be the value specified in the builtin az provider version.
```
anthony-c-martin, line 193: > also not intuitive

This is an opinion. You could argue that "the closest bicepconfig.json is used" is intuitive and simpler to reason about.

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,259 @@
...
+
+```json
+{
+  // prior bicepconfig.json sections
+  // ...
+
+  "providers": {
+    "az": {
+      // A package reference string (below) gestures Bicep to source the types from the registry address.
+      "source": "mcr.microsoft.com/bicep/providers/az:0.2.3",
```
alex-frankel, line 96: In the 99% case, types are going to be pulled from the public registry. Can we provide an option for only adjusting the version of az instead of the full or aliased registry path?

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,247 @@
...
+
+In this case Bicep will try to resolve the `modFooProvider` details from `src/modules/bicepconfig.json`. The config doesn't specify a `providers` section, and the builtin config doesn't include a definition for provider `foo` and so the compiler cannot resolve the provider declaration statement and will emit a diagnostic.
+
+If we replace `foo` with `az` in the example above, Bicep resolves `modFooProvider`'s version to be the version of `az` specified in the builtin configuration and not the version `1.2.3` specified in `src/modules/mod.bicep` (inherited from `src/bicepconfig.json`) which is also not intuitive. The reson for this behavior is that current `bicepconfig.json` merge semantics merge `src/modules/bicepconfig.json` with the builtin config and the version applied would be the value specified in the builtin az provider version.
+
+This is especially problematic for 3rd party namespace identifiers specified in a `bicepconfig.json` up a file hierarchy, since they would result in a compilation error because they are unspecified in the builtin configuration.
+
+1. TODO: Potential drawback when trying to inspect module sources 
+<!--
```
anthony-c-martin, line 198: It would be good to get more clarity about the intention of the "inspect module sources" feature. If the goal is for the original source to be represented in full fidelity and compilable, then I agree this is important. If it's just to provide a quick reference (similar to firing up github or ADO code search), then this doesn't feel as important.

This is also technically orthogonal, as it's a problem regardless.

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,247 @@
...
+      },
+      "isDefaultImport": true
+    }
+  }
+}
+-->
+
+## Alternatives
+
+- Using a separate file other than `bicepconfig.json` to store resource type provider version information was evaluated but adds complexity to UX as users need to be aware of two separate merge semantics (one for config and one for provider packages)
```
anthony-c-martin, line 243: Forcing users to maintain two similar-but-different configuration files adds complexity. IMO we should try to avoid this at all costs.

active/0002-centalized-mgmt-resource-types-provider-versions.md
jeskew: The best prior art on merging that I recall from previous discussion is [the `"extends"` property in tsconfig.json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#tsconfig-bases). NPM doesn't perform any kind of merge, though there are some tools that let you do merging (like Lerna).

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: This design proposal doesn't preclude us from implementing wildcard semantics in the provider version. 
We will add the question of whether this is a hard requirement for GA or can be an incremental feature post GA.

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: Let's look out for costumers' feedback on the first iteration and discuss this as part of GA release

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: TODO(asilverman): Move this discussion to the out-of-scope section

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,259 @@
+---
```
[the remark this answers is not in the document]
asilverman: yes, I was surprised by it as well but its unrelated to this REP, I recommend you create a task for someone to fix

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: rephrased to remove the example of package.json as I learned that npm doesn't support merge semantics natively

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: replaced intuitive with expected

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: I decided to reduce duplicate names and simplify this to the following:
```json
{
    "providers":{
        "az":{
            "source":  "mcr.microsoft.com/bicep/providers",
            "version: "1.2.3",
            "isDefaultImport": true
        }
    }
}
```

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
...
+
+- [Updating `bicepconfig.json` merge semantics](#updating-bicepconfigjson-merge-semantics)
+- [Resolution of namespace identifiers by VSCode extension when inspecting published module sources](#resolution-of-namespace-identifiers-by-vscode-extension-when-inspecting-published-module-sources)
+- [Wildcard semantics for provider versions in `bicepconfig.json`](#wildcard-semantics-for-provider-versions-in-bicepconfigjson)
+
+## Detailed design
+
+### Client side changes
+
+#### Add a new syntax for the provider import declaration statements: `import {providerName} [as {optionalAlias}]`
```
jeskew, line 65:
```suggestion
#### Add a new syntax for the provider import declaration statements: `provider {providerName} [as {optionalAlias}]`
```

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+#### Change the delimiter '@' used by the supported provider declaration syntax to ':' to align with module reference syntax and increase consistency
+
+The syntax shown below uses ':' instead of '@' as a delimiter for provider declaration statements
+
+```bicep
+provider 'br:mcr.microsoft.com/bicep/providers/az:0.2.3' // notice its NOT using az@0.2.3
+provider 'br/public:az:0.2.3' as myAz
+```
```
jeskew, line 122: Is this section still relevant? Two sections up, the proposal is to move registry and versioning info from `provider` statements to bicepconfig

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Drawbacks
+
+- Costumers may be confused by the config merge semantics behavior
+- Costumers will not be able to tell from inspecting sources alone what is the provider version used to resolve resource types
+
+## Alternatives
```
jeskew, line 129:
```suggestion
## Alternatives considered
```

[Very minor, totally optional]: "Alternatives" to me implies that what is listed below is still a viable option. "Alternatives considered," however, implies that you thought the following through but decided not to do it. There's an implied "and rejected" following "Alternatives considered"

jeskew (commented):
Just had a couple nitpicky comments.

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: It is still relevant since the syntax will continue to be accepted. The addition of centralized provider version management is not exclusive with the syntax highlighted above.

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: Yeah I think that works, however I think likely this should be a change made in the template. The REP process delineated is to follow the template and this is what I did. I think your thoughts here are valid and should be discussed as a separate issue on the REP process

active/0002-centalized-mgmt-resource-types-provider-versions.md
jeskew: ... is the plan to keep both all the way through GA of this feature? [You make some pretty convincing arguments above as to why we shouldn't allow direct registry references inline](https://github.com/Azure/bicep-reps/pull/2/files#diff-1c3e322d927a938d42051e08eb8e935b46bd1fbfc6360cc11cb63ecab3df2914R48), and supporting both inline and configuration-driven provider sourcing seems like it would result in the maximum possible complexity (both in terms of code complexity and in terms of the complexity for users adopting this feature).

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
+---
+REP Number: 0002
+Author: asilverman (Ariel Silverman)
+Start Date: 2023-12-19
+Feature Status: Private Preview
+Bicep Issue Number(s): [12202](https://github.com/Azure/bicep/issues/12202)
+---
+
+# Title - Centralized management of resource type provider versions using package.json
```
alex-frankel, line 9: nit - this still says `package.json` but I think you mean `bicepconfig.json`?

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
...
+
+The resolution of the symbols will be resolved from the builtin `bicepconfig.json` under a new section:
+
+```json
+{
+  // prior bicepconfig.json sections
+  // ...
+
+  "providers": {
+    "az": {
```
alex-frankel, line 82: are the dictionary keys going to be validated? In other words, if `az` does not exist at the location `mcr.microsoft.com/bicep/providers` will an error be thrown?

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
...
+
+```bicep
+provider 'br:mcr.microsoft.com/bicep/providers/az@1.2.3' // fully-qualified
+provider 'br/public:az@1.2.3' // uses provider aliasing defined in bicepconfig.json
+```
+
+We also currently support the following syntax for "built-in" providers:
+
+```bicep
+provider 'sys@1.0.0'
```
shenglol, line 34: I would like to point out that `sys` is more like a compile-time namespace import (the feature Jonny has been working on), not a provider, since all it contains are some built-in functions. We used to share the keyword `import` between providers and imports, but now that we changed the keyword for providers, at some point we will need to update the implementation for `sys` to make it a compile-time import. This is out of scope of this REP though.

anthony-c-martin: Conceptually I think there's overlap with both imports & providers (at some point we will add function execution to providers, and "az" has functions). This would be an interesting discussion topic, but agree we can consider it out-of-scope for now.

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
...
+      "isDefaultImport": false
+    }
+  }
+}
+```
+
+Since previous provider declaration syntax forms continue to be supported, we introduce the following constraints:
+
+- To ensure consistency with pre-existing handling of `bicepconfig.json` the configuration file closest to the Bicep file in the directory hierarchy is used.
+- The `sys` namespace is coupled to the Bicep bits version and cannot be overriden or dynamically uploaded using the mechanism described above
```
anthony-c-martin, line 104: Could you clarify the editor behavior if a user tries to define an entry named "sys" in this dictionary? Would we treat this as if the config file were invalid?

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
...
+
+#### Remove support for syntax `import '{providerName}@{providerVersion}' [as {optionalAlias}]`
```
anthony-c-martin, line 106:
```suggestion
#### Remove support for syntax `provider '{providerName}@{providerVersion}' [as {optionalAlias}]`
```

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
...
+1. The new syntax is quite verbose, and cannot be controlled centrally. For users with a large set of files, picking up the latest version of `az` will require modifying every single file.
+1. The aliasing (assuming that the last segment of the registry namespace == the provider name) & versioning (having an `@` character instead of a `:` which is familiar to module users) behavior of the new syntax can be confusing.
+1. Its deceiving and confusing to allow `sys` to have a pinned versionis fixed version `1.0.0` - this **should** imply that there are no changes between Bicep versions, which isn't realistic.
+
+- It's an "either/or" - either users follow the current behavior (`az` imported by default) **or** they're using the new syntax. We don't have a plan for how to combine the two.
+- With [Radius](https://github.com/radius-project/radius), it's conceivable that we may want users to **not** have `az` imported by default.
+
+### Proposed Changes
+
+- [Updating `bicepconfig.json` merge semantics](#updating-bicepconfigjson-merge-semantics)
```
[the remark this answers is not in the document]
asilverman: Its out of scope, thanks for calling this out it is an oversight on my part

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: Added a call out in the out of scope section regarding this topic.

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: During the compilation process if the artifact reference `mcr.microsoft.com/bicep/providers/az:0.2.3` cannot be resolved for any reason, the system will emit a diagnostic with a reasonable error message that informs the user regarding the nature of the failure if one is faound.

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: yes, thank you. Fixed now

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: The current plan of record is to support all three mechanisms of specifying a provider. The conversation on whether this should be the only and exclusive way to do it is not in scope for this proposal but it is a valid question and we can have a separate discussion about it. @anthony-c-martin, can you please add your thoughts on this?

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: I can think of two ways to handle this scenario:

1. Forbid the use of the value `sys` as an entry in the dictionary under that section of the `bicepconfig.json`.
2. Allow this to be specified but emit a diagnostic when the author writes the provider declaration statement `provider sys`
   the diagnostic would be `BCP084: The symbolic name "sys" is reserved. Please use a different symbolic name. Reserved namespaces are "az", "sys".`

   The easy one to reason from a customers perspective is the first one in my opinion. I am open to other potential resolutions if my choice is not acceptable.

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,262 @@
...
+{
+  "providers": {
+    "az": {
+      "source": {
+        "registry": "mcr.microsoft.com/bicep/providers",
+        "version": "0.2.3"
+      },
+      "isDefaultImport": true
+    },
+    "sys": {
```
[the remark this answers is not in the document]
asilverman: alright, I think we can do this and postpone the decision of whether its a provider or an import. I chose to frobid specifying it as a key in that dictionary.

active/0002-centalized-mgmt-resource-types-provider-versions.md
alex-frankel: FWIW, I would be ok if we started with only the centralized method via bicepconfig and then expanding as needed/requested.

alex-frankel (approved):

active/0002-centalized-mgmt-resource-types-provider-versions.md
shenglol: +1. I think maintaining multiple methods of specifying provider versions and later removing certain mechanisms could result in breaking changes.

active/0002-centalized-mgmt-resource-types-provider-versions.md
anthony-c-martin: First option sounds good - could you add that to the rep?

active/0002-centalized-mgmt-resource-types-provider-versions.md
anthony-c-martin: This discussion on whether or not we want an in-line mechanism feels orthogonal to this spec. Could we agree to what @asilverman is proposing (change the delimiter) and plan to have a follow-up discussion on whether or not we should support in-line overrides generally?

jeskew: Fair... I had just assumed this section had just been accidentally left in from a previous version since my impression from the team discussion of this proposal was that the inline mechanism would be replaced. But that does seem like something we can deal with while this proposal is still active.

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,277 @@
...
+      "source": {
+        "registry": "mcr.microsoft.com/bicep/providers",
+        "version": "0.2.3",
+      },
+      "isDefaultImport": true // Gestures Bicep to implicitly import the namespace identifier to the global scope
+    },
+    "kubernetes": {
+      // Using the value "builtin" (below) gestures Bicep to source types from the statically compiled NuGet package reference
+      "source": {
+        "builtin": true
```
majastrz, line 92: Should we change the casing to `builtIn` since this is equivalent to built-in in normal English.

asilverman: yeah, that works

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,277 @@
...
+bicepconfig.json
+
+```jsonc
+{
+  "providers": {
+    "az": {
+      "source": {
+        "registry": "mcr.microsoft.com/bicep/providers",
+        "version": "0.2.3"
+      },
```
shenglol, line 246: Should we think about how to handle conflicts? Two potential scenarios come to mind:

The user tries to import the az or kubernetes provider from a private ACR.
The user tries to import a third-party provider that happens to be named az or kubernetes 😈, but the provider is not the built-in az or kubernetes provider we own. While we currently don't support third-party providers, it could be beneficial to explore potential considerations for such scenarios.

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,277 @@
...
+      "isDefaultImport": true
+    }
+  }
```
shenglol, line 249: Should we consider how to handle conflicts? Two scenarios come to mind:

- The user tries to import `az` from both MCR and ACR.
- The user tries to import a third-party provider whose name happens to be `az` 😈. Although we don't support third-party providers today, it might be worth thinking through the design.

asilverman: Thanks for these comments, they are on point and useful: 
On the first point I made a change to fully qualify the repository in the `sources` key, also, added a constraint so that all keys in the `providers` object are unique. I believe this solves the issue as the user is now able to import both provider packages and name their identifier without name collision.
The second scenario seems to be an instance of the same limitation that was addressed by the changes mentioned above.

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -274,4 +251,8 @@ At this time we are not proposing support for wildcard semantics on provider ver
...
\ No newline at end of file
+It is possible to think of `sys` more like a compile-time namespace import rather than a provider, since all it contains are some built-in functions. We used to share the keyword import between providers and imports, but now that we changed the keyword for providers, at some point we will need to update the implementation for sys to make it a compile-time import. This is out of scope of this REP however.
+
+### Implicit import of provider identifiers to global scope
+
+In the current impementation two provider identifiers are implicitly imported: `sys` and `az`, their types are available in the global scope and the author doesn't neeed to dereference types using the identifier (i.e. `resourceGroup() vs az.resourceGroup()`). A prior version of this proposal introduced the concept of `importProviderByDefault` to allow users to appy the aforementioned behavior on other providers. This is out of scope for this proposal and is not advised since pollution of the global namespace makes the code harder to read.
```
anthony-c-martin, line 258: How would a user opt-out of `az` being imported by default without this field?

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -274,4 +251,8 @@ At this time we are not proposing support for wildcard semantics on provider ver
 
 ### Deciding if `sys` is a compile time import or a resource types provider package
 
-It is possible to think of `sys` more like a compile-time namespace import rather than a provider, since all it contains are some built-in functions. We used to share the keyword import between providers and imports, but now that we changed the keyword for providers, at some point we will need to update the implementation for sys to make it a compile-time import. This is out of scope of this REP however.
\ No newline at end of file
+It is possible to think of `sys` more like a compile-time namespace import rather than a provider, since all it contains are some built-in functions. We used to share the keyword import between providers and imports, but now that we changed the keyword for providers, at some point we will need to update the implementation for sys to make it a compile-time import. This is out of scope of this REP however.
```
anthony-c-martin, line 254: I don't follow - why is `sys` more like a compile-time import than a provider? My impression is the complete opposite.

Compile-time imports require referencing variables or functions from another Bicep/JSON file. Providers allow referencing behavior from the runtime.

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,258 @@
...
+1. The aliasing (assuming that the last segment of the registry namespace == the provider name) & versioning (having an `@` character instead of a `:` which is familiar to module users) behavior of the new syntax can be confusing.
+1. Its deceiving and confusing to allow `sys` to have a pinned version fixed to `1.0.0` - this **should** imply that there are no changes between Bicep versions, which isn't realistic.
+
+- It's an "either/or" - either users follow the current behavior (`az` imported by default) **or** they're using the new syntax. We don't have a plan for how to combine the two.
+- With [Radius](https://github.com/radius-project/radius), it's conceivable that we may want users to **not** have `az` imported by default.
+
+### Proposed Changes
+
+- [Resolution of namespace identifiers by VSCode extension when inspecting published module sources](#resolution-of-namespace-identifiers-by-vscode-extension-when-inspecting-published-module-sources)
+- [Wildcard semantics for provider versions in `bicepconfig.json`](#wildcard-semantics-for-provider-versions-in-bicepconfigjson)
```
anthony-c-martin, line 58: Both of these link to headings under "out of scope". Are they out of scope or in scope?

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,258 @@
...
+The syntax shown below uses ':' instead of '@' as a delimiter for provider declaration statements
+
+```bicep
+provider 'br:mcr.microsoft.com/bicep/providers/az:0.2.3' // notice its NOT using az@0.2.3
+provider 'br/public:az:0.2.3' as myAz
+```
+
+## Drawbacks
+
+- Costumers may be confused by the configt merge semantics behavior
```
anthony-c-martin, line 122: [nit] Costumers -> Customers here and in multiple other places. I also generally prefer to just say "users" rather than "customers", as I feel like it's more appropriate for an open-source project (less "us" vs "them")

active/0002-centalized-mgmt-resource-types-provider-versions.md
```diff
@@ -0,0 +1,258 @@
...
+The current syntax (shown above) presents several challenges:
+
+1. The new syntax is quite verbose, and cannot be controlled centrally. For users with a large set of files, picking up the latest version of `az` will require modifying every single file.
+1. The aliasing (assuming that the last segment of the registry namespace == the provider name) & versioning (having an `@` character instead of a `:` which is familiar to module users) behavior of the new syntax can be confusing.
+1. Its deceiving and confusing to allow `sys` to have a pinned version fixed to `1.0.0` - this **should** imply that there are no changes between Bicep versions, which isn't realistic.
+
+- It's an "either/or" - either users follow the current behavior (`az` imported by default) **or** they're using the new syntax. We don't have a plan for how to combine the two.
+- With [Radius](https://github.com/radius-project/radius), it's conceivable that we may want users to **not** have `az` imported by default.
+
+### Proposed Changes
```
anthony-c-martin, line 55: [nit] Should this be a level 2 heading (`##`) rather than 3?

active/0002-centalized-mgmt-resource-types-provider-versions.md
anthony-c-martin: As discussed today, I'm good with special casing "sys", and blocking it in the bicepconfig. If we're considering it to not be a provider, should we also block the following?

```bicep
provider sys
```

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: things got scrambled here, fixing now...

active/0002-centalized-mgmt-resource-types-provider-versions.md
asilverman: I think since the conversation is still not settled, lets just treat it as any other provider and call out in the docs what are the consequences of doing this so users dont end up shooting themselves in the foot

anthony-c-martin: > I think since the conversation is still not settled, lets just treat it as any other provider and call out in the docs what are the consequences of doing this so users dont end up shooting themselves in the foot

I thought we settled this in the discussion today? Here's my recollection:
1. We do not consider "sys" to be a provider - it is instead an intrinsic feature of the runtime. So regular "provider" semantics don't need to apply to it, and we expect this to fit with the user's mental model of what a provider is.
1. To prevent weird behavior, we should block "sys" in the "providers" configuration section introduced by this rep.
1. "sys" should not appear or be required in the "implicitProviders" config we discussed.

We didn't discuss this today, but to me the following is also necessary to be consistent with the above:
1. We should block usage of the syntax `provider sys`

If we're happy with all of the above, then I think we can document it in this rep and move on.

anthony-c-martin (approved):

PR #3, opened 2024-01-25
Extensible Resource Deployment Observability Improvements

shenglol:


active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,226 @@
...
+    ]
+  }
+}
+```
+
+#### Updating the `outputResources` to include more identifiable information
+
+To improve the current structure where each `outputResources` object only has a resource `id`, it's essential to add more properties for identifying extensible resources, which lack resource IDs. These additions should include the symbolic name, the resource type, and the deployment provider alias.
+
+```diff
```
majastrz, line 96: I had no idea there was a `diff` "language" you could use in md files!

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,226 @@
...
+{
+  ...
+  "type": "Microsoft.Resources/deployments",
+  "properties": {
+    "outputResources": [
```
majastrz, line 101: `outputResources` excludes `existing` resources today. We're not planning to change that?

shenglol: Let me add this to open questions and discuss it further. It should be a simple change implementation-wise, but if we include `existing` resources, the meaning of `outputResources` is changed. Personally, I'd prefer adding a `referencedResources` for that purpose.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,230 @@
...
+          {
+            "resourceType": "storageAccounts",
+            "locations": [
+              "eastus"
+            ]
+          }
+        ]
+      }
+    ],
++   "deploymentProviders": [
```
anthony-c-martin, line 76: "deployment" & "providers" are both pretty heavily overloaded and commonly used. What about something like "importedProviders"?

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,230 @@
...
++       "version": "1.27.8"
++     }
+    ]
+  }
+}
+```
+
+#### Updating the `outputResources` to include more identifiable information
+
+To improve the current structure where each `outputResources` object only has a resource `id`, it's essential to add more properties for identifying extensible resources, which lack resource IDs. These additions should include the symbolic name, the resource type, and the deployment provider alias.
```
anthony-c-martin, line 94: > identifying extensible resources, which lack resource IDs

Doesn't the v2 contract change this?

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,230 @@
...
+
+```diff
+{
+  ...
+  "type": "Microsoft.Resources/deployments",
+  "properties": {
+    "outputResources": [
+      // Azure resource
+      {
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Storage/storageAccounts/my-storage-account"
```
anthony-c-martin, line 104: What about azure resources using symbolic names? Wondering if it's useful to also capture symbolic names for these resources.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,230 @@
...
+      },
+      // Extensible resource
+      {
```
anthony-c-martin, line 107: AFAIK `outputResources` today captures resources from nested deployments as well. This works because the `resourceId` is globally unique within a tenant. The same is not true for symbolic names - do we need something to identify which nested deployment they relate to?

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,230 @@
...
+
+Implementing this enhancement involves comprehensive modifications across various deployment engine components, including but not limited to `AzureDeploymentEngine`, `ResourceProxyDefinition`, `DeploymentOutputResourceDefinition`, and numerous deployment jobs. Such widespread changes carry a significant risk due to the complexity of the deployment engine and the potential for unforeseen impacts on its functionality, which would require careful consideration and discussion.
+
+### Should the az provider always be returned?
+
+Building on the previous discussion, if the deploymentProvider details are not included within `outputResources` for Azure resources, should we consistently list the `az` provider in the `deploymentProviders` for sake of consistentcy and clarity?
+
+### Should `outputResources` include `existing` resources as well?
+
+Currently, `outputResources` only contains created or updated resources. Should we expand the property to include `existing` resources?
```
anthony-c-martin, line 223: To answer this question, I think it would be good to answer the question of "What do we expect people to do with the `outputResources` property?".

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Out of scope
+
+### Portal UI design
+
+The update to the Portal deployments blade is necessary to display the symbolic name and deployment provider information for extensible resources. Since the portal team has ownership of the deployments blade, they hold the responsibility for both the UI design and the implementation of these changes.
```
anthony-c-martin, line 229: I think the minimum we'd need is to make sure we give them a clear way to differentiate between an "Azure" and an extensible resource programmatically. I think you've covered that in this spec, but it might be worth stating it explicitly.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: Makes sense. Added a comment to mention this as an alternative, and added it as an unresolved question. Personally, I don't have a strong opinion on the property name, and I think we can do a vote during next week's Bicep discussion.

Sharing some other alternatives suggested by ChatGPT:
- infraProviders
- infrastructureProviders
- platformProviders
- platforms

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: Yeah, I'm going to mention it in a follow-up REP to be sent out shortly. One question I have is whether we should reuse the property name `id`, as some clients may expect it always to be a valid resource ID, but for extensible resources ID is a freeform string.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: That has some challenges in terms of implementation. This was already captured as an unresolved question below.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: For context, orginally I wrote a single REP that covered both this feature and the stacks integration. However, as the document expanded, it became increasingly complex and challenging to navigate, given it essentially addressed two topics, so I decided to separate it into two for clarity.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: Good catch, I missed that part. It seems we would need to include the deployment name as well.

shenglol: We may also need to add deployment name to `deploymentProviders` to disambiguate 🤔.

shenglol: Added a `deploymentId` property to resolve this.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: Makes sense. Updated this section to mention it explicitly.

active/0003-extensible-resource-deployment-observability-improvements.md
majastrz: Oh definitely. I mainly wanted to confirm the behavior isn't changing since Stacks takes a dependency on that.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,249 @@
...
+### Deployment Provider
+
+An API abstraction for the following operations:
+
+- Creating or updating Azure Resources or extensible resources declared in Bicep files or an ARM templates;
+- Performing resource-level or provider-level queries that have no side effects through provider functions in Bicep files or ARM templates, e.g., list secrets of a resource (currently, this is only supported by the az provider).
+
+### Extensibility Provider
+
+An extensibility provider refers to any deployment provider that operates outside the scope of ARM, such as the MS Graph provider and the Kubernetes provider. An extensibility provider deployes extensible resources instead of Azure resources.
```
asilverman, line 25: nit:
> An extensibility provider refers to any deployment provider that operates outside the scope of ARM, such as the MS Graph provider and the Kubernetes provider. An extensibility provider **deploys** extensible resources instead of Azure resources.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,249 @@
...
+
+This REP outlines API enhancements for [`Microsoft.Resources/deployments`](https://learn.microsoft.com/en-us/rest/api/resources/deployments/create-or-update?view=rest-resources-2021-04-01&tabs=HTTP) and [`Microsoft.Resources/deployments/operations`](https://learn.microsoft.com/en-us/rest/api/resources/deployment-operations/get-at-scope?view=rest-resources-2021-04-01) to empower users to effectively monitor the progress of extensible resource deployments through Azure CLI, Azure Powershell, and Azure Portal deployments blade, and other relavant Azure Resource Manager interfaces.
+
+## Terms and definitions
+
+### Deployment Provider
+
+An API abstraction for the following operations:
+
+- Creating or updating Azure Resources or extensible resources declared in Bicep files or an ARM templates;
```
stephaniezyen, line 20: - Creating or updating Azure Resources or extensible resources declared in Bicep files or ARM templates;
Just a tiny grammar nitpick - to take out "an" before ARM Templates

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,249 @@
...
+      {
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Storage/storageAccounts/my-storage-account"
+      },
+      // Extensible resource
+      {
++       "symbolicName": "myService",
++       "resourceType": "core/Service@v1",
++       "deploymentId": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Resources/deployments/my-deployment",
++       "deploymentProvider": "kubernetes"
+      }
```
rynowak, line 113: How would a user or tool identify the actual resource? This is the difference between:

- *A Kubernetes `Deployment`* was created
- *A Kubernetes `Deployment` called `my-deployment` in namespace `my-namespace` was created

I don't think the former is super useful. The symbolic name doesn't help because it refers to the implementation details of the template (which can change).

active/0003-extensible-resource-deployment-observability-improvements.md
anthony-c-martin: Wanted to revisit this in the discussion today. It feels like an "id" would be more useful for the purposes we use the `outputResources` field today (tracking changes to specific resources). Having the symbolic name would be useful for user understanding, but doesn't feel like it solves the same problem.

I think @rynowak's comment is related.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,249 @@
...
+
+```
+VERBOSE: 00:00:00 - Extensible resource with symbolic name 'myService' of type 'core/Service@v1' has provisioning status: 'Succeeded'. The resource is provisioned by 'Kubernetes@1.27.8'.
+```
+
+## Drawbacks
+
+### Inconsistency between Azure resources and extensible resources
+
+The new properties are added exclusively for extensible resources, leaving Azure resources unchanged, which avoids introducing breaking changes to the API, SDK, and clients. However, this approach leads to an inconsistency in the schema between Azure resources and extensible resources. Such a disparity could potentially cause confusion, as users might expect a uniform schema across all resource kinds. Clear documentation might be required to manage user expectations and explain the rationale behind the differing schemas.
```
rynowak, line 158: You could add the same fields with the same meaning for Azure. Imagine if every output resource defined:

- `symbolicName`
- `resourceType`
- `deploymentId`
- `deploymentProvider`
- (additional *keys* as necessary to identify the resource, for Azure this is `id`)

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,249 @@
...
+
+> It's worth noting that we cannot use the name `providers` as the property `providers` is already assigned to return Azure Resource Provider information. While an alternative approach could involve renaming the current `providers` property to `resourceProviders` for enhanced accuracy, the solution should be avoided as it constitutes a breaking change.
+
+```diff
+{
+  ...
+  "type": "Microsoft.Resources/deployments",
+  "properties": {
+    ...
+    "providers": [
```
asilverman, line 61: What's the consideration to create a new array vs append to this array?

active/0003-extensible-resource-deployment-observability-improvements.md
rynowak: We definitely have a usecase for this in Radius.

active/0003-extensible-resource-deployment-observability-improvements.md
anthony-c-martin: This may be too much of a breaking change to justify, but another option that might be worth discussing is having the `outputResources` just return resources deployed directly, and to require the caller (i.e. Stacks) to traverse the deployment graph.

Re-opening to discuss today.

active/0003-extensible-resource-deployment-observability-improvements.md
rynowak: This is one of the places where our fork of the DE deviated from the official one. We wanted to know which resources were *used* in a deployment and which ones were *mutated* by a deployment. 

This is information that Radius catalogs to help describe the application architecture.

shenglol: We discussed the issue today and decided to not include existing resources in `outputResources` since it will be a breaking change. To support Radius's use case, I think we can add a new property like `referencedResources`.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: Appending to this array means we will be mixing "Azure resource providers" with "extensibility providers", which I think we should avoid due to their conceptually distinct natures.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: Missed this in today's meeting, but I guess we can discuss it asynchronously here. Restricting `outputResources` to only include resources deployed in the current deployment can make the API more concise and relevant to the specific deployment context. Do we know when `outputResources` is introduced? If it's a longstanding property not originally designed for Deployment Stacks, making the breaking change is a worrying, since removing elements from an established array is a kind of behavioral change that might not be immediately apparent to users.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: For users, `deploymentId`` and `symbolicName`` should suffice to link a resource to its Bicep module and its declaration within that module, assuming the template remains unchanged. However, for services like Deployment Stacks that need to uniquely identify a resource even when the template changes, a unique ID for extensible resources is necessary. This concept is further explored in the subsequent REP: https://github.com/Azure/bicep-reps/pull/4.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: Indeed, this is actaully mentioned in the "Unresolved Questions" section below. For the sake of consistency, implementing this seems appropriate. However, after discussion with other team members yesterday, the decision was made to refrain from implementation at this time. This is due to the fact that integrating the additional properties for Azure resources would necessitate changes across quite a few components of the deployment engine, which carries a considerable risk. Moreover, given that Deployment Stacks might be the sole consumer the `outputResources`, the effort required may not justify the benefits.

active/0003-extensible-resource-deployment-observability-improvements.md
shenglol: Good catch. Fixed the typo.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,257 @@
...
+- `managementProviders`
+- `apiProviders`
+- `platformProviders`
+- `platforms`
+- `apis`
+- `plugins`
+- `vendors`
+- `protocols`
+- `services`
+- `workloads`
```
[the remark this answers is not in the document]
shenglol: Had an offline discussion on this. The decision is to proceed with the term `extensions`. Will update the REP to incorporate the change.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,280 @@
...
+  "properties": {
+    "outputResources": [
+      // Azure resource
+      {
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Storage/storageAccounts/my-storage-account"
+      },
+      // Extensible resource
+      {
++       "deploymentId": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Resources/deployments/my-deployment",
++       "extension": "kubernetes"
```
kalbert312, line 110:
```suggestion
+       "extensionAlias": "kubernetes"
```

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,280 @@
...
+  "properties": {
+    "outputResources": [
+      // Azure resource
+      {
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Storage/storageAccounts/my-storage-account"
++       "symbolicName": "myStorageAccount",
++       "resourceType": "Microsoft.Storage/storageAccounts",
++       "apiVersion": "2020-01-01",
++       "deploymentId": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Resources/deployments/my-deployment",
++       "extension": "az"
```
kalbert312, line 223:
```suggestion
+       "extensionAlias": "az"
```

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,280 @@
...
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Storage/storageAccounts/my-storage-account"
+      },
+      // Extensible resource
+      {
++       "deploymentId": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Resources/deployments/my-deployment",
++       "extension": "kubernetes"
++       "symbolicName": "myService",
++       "resourceType": "core/Service",
++       "apiVersion": "v1",
++       "identifiers": {
```
kalbert312, line 114: I will disambiguate ARM vs EXT resources by the existence of the `id`\`identifiers` property just in case a future API update includes the other properties for Azure resources. Not sure if a `isExtensionResource` property is worth it.

shenglol: `extension` is required to uniquely identify extensible resources.

shenglol:
Kyle is now driving the design and development as part of the stacks and extensibility integration efforts.

PR #4, opened 2024-01-27
Integrating extensible resources and deployment stacks

shenglol:


active/0004-integrating-extensible-resources-and-deployment-stacks.md
```diff
@@ -0,0 +1,547 @@
...
+
+Implementing these changes ensure that extensible resources can be managed with the same level of efficiency and security as Azure resources within Deployment Stacks.
+
+## Detailed design
+
+### Reference ID
+
+> A reference ID is a URL-encoded, free-form string designed to uniquely identify an extensible resource. This ID is generated by extensibility providers and is returned as part of the responses from the `POST /resources/reference` or `POST /resources/preview` extensible resource APIs. The concept and design of reference IDs have been finalized.
+>
+> Any reference ID MUST adhere to the format `{flags}-{uniqueId}`.
```
rynowak, line 41: Is there a technical reason why we need to encode two values in one field? Would something break if the flags were separated from the unique id in two fields?

active/0004-integrating-extensible-resources-and-deployment-stacks.md
```diff
@@ -0,0 +1,547 @@
...
+
+> - `{flags}`: This section consists of two binary digits indicating the type of reference ID. The currently accepted values are `00` and `01`:
+>   - `00`: signifies that the reference ID is autonomous and doesn't need extra data for GET and DELETE actions. This implies the utilization of the extensible resource APIs `GET /resources/{referenceId}` and `DELETE /resources/{referenceId}` for operations.
```
rynowak, line 44: I'm curious, and I might have missed a previous design session on the topic...

What's the difference between a provider that *defines an empty configuration* and a provider that *doesn't require configuration*?

active/0004-integrating-extensible-resources-and-deployment-stacks.md
```diff
@@ -0,0 +1,547 @@
...
+>   - `01`: indicates that additional information, specifically user-provided provider configuration (like a connection string), is essential alongside the reference ID to execute GET and DELETE operations. In this case, APIs `POST /resources/{referenceId}/getWithConfig` and `POST /resources/{referenceId}/deleteWithConfig` must be used, and the provider configuration must be included in the request body.
```
rynowak, line 45: Apologies if I missed a previous discussion on this ...

Have you considered turning this into a more RPC-style API? The trailing operation name (eg: `/deleteWithConfig` at the end of the URL path) is hard to parse in an unambiguous way. 

ARM has rules to disambiguate this, but I'm not sure this will be true of all reference ID formats. In any case where the reference ID has a variable number of segments, this is pretty difficult to handle, and your web framework can't help.

For an example in ASP.NET core, you can write a route to handle `/resources/{**referenceId}`, but you can't write a route to handle `/resources/{**referenceId}/deleteWithConfig` (it's blocked by the routing engine). Most routing frameworks have the same limitation.

active/0004-integrating-extensible-resources-and-deployment-stacks.md
```diff
@@ -0,0 +1,547 @@
...
+> - `{uniqueId}`: This is a URL-encoded free-form string that serves to uniquely identify a resource. The format of uniqueId is defined by the providers and is opaque to the Azure Deployments engine.
+>
```
rynowak, line 47: It looks like the reference ID is passed in the URL - not the unique ID. If my unique ID for a resource is "a-cool-resource" then the requests would look like:

- `POST /resources/00-a-cool-resource`
- `DELETE /resources/01-a-cool-resource/deleteWithConfig`

It feels like the "flags" in this case are really serving as an API-version.

active/0004-integrating-extensible-resources-and-deployment-stacks.md
```diff
@@ -0,0 +1,547 @@
...
+
+To overcome these limitations, the REP suggests modifications to the deployments API. A proposed change is the introduction of a `referenceID` field in `outputResources`, a freefrom string designed to uniquely identify extensible resources. Additionally, the proposal includes updates to the `deploymentProviders`, enabling it to store provider configurations. These configurations may include references to key vaults or resource properties, which can be utilized to access authentication details safely. These proposed enhancements allows Deployment Stacks to interact with extensibility providers to delete extensible resources.
+
+Implementing these changes ensure that extensible resources can be managed with the same level of efficiency and security as Azure resources within Deployment Stacks.
+
+## Detailed design
+
+### Reference ID
+
+> A reference ID is a URL-encoded, free-form string designed to uniquely identify an extensible resource. This ID is generated by extensibility providers and is returned as part of the responses from the `POST /resources/reference` or `POST /resources/preview` extensible resource APIs. The concept and design of reference IDs have been finalized.
```
rynowak, line 39: I question whether it's the right thing to **store** the URL-encoded form. Can you explain more about why?

It feels like URL-encoding should be done by the client rather than part of the representation. Most client libraries take a *secure-by-default* approach and assume that data supplied by the caller needs-to-be-encoded. You usually need to do extra work in a client program to supply an already-encoded URL. 

This can lead to subtle bugs when a reference ID can, but rarely will contain URL-unsafe characters. Most of the time the code will work.

active/0004-integrating-extensible-resources-and-deployment-stacks.md
shenglol: Is this true if `referenceId` even if it is URL-encoded?

rynowak: If the API call is going to include a URL-encoded string **inside** a path segment like `/resources/<url-encoded-text-here>/deleteWithConfig` then you're going to see inconsistent behaviors across webservers when that segment contains `%2F` (encoded text for `/`) or path navigation operators like `..` which don't require encoding.
 
For example I believe ASP.NET Core (using it as an example because I know it well) behaves differently depending on whether IIS is in front of it or not.
 
Servers take differing approaches to cases like these because URL paths often map to authorization rules and other security features. The risk that they are mitigating is when the server code (or middleware) interprets the URL differently than application code.
 
If uniqueId is a free-form string then passing it in the request body is the safest option. This is already an RPC-style API, so I'm not sure why it's valuable to pass the uniqueId in the URL.
 
Let me know how I can help, this design feels like it could be simplified.

shenglol: This is good to know! The other comments and questions also make sense. As discussed offline, the v2 APIs require a revision. Will update this PR later to incorporate the changes once the [new extensibility v2 API design](https://github.com/Azure/bicep-extensibility/pull/176) is approved.

shenglol:
The design is now driven internally by the Stacks team. Closing the REP.

PR #6, opened 2024-08-27
Integration of extension resources with deployment stacks.

kalbert312:


active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+Allowed expressions:
+- ARM resource access that can be compiled to a resource ID (`aksCluster` in the example).
+- ARM resource API calls that can be compiled to a URI, HTTP method, and any inputs (`listClusterAdminCredential` in the example).
+- JSON value accessors (`kubeconfig[0].value` in the example).
+- Usage of compile-time constants (the `0` in `kubeconfig[0]`, `namespace: 'default'` in the example).
+
+Disallowed expressions:
+- Usage of deployment parameters.
+- Usage of Bicep or ARM template functions (built-in or user defined).
+- Usage of branched expressions (ternaries).
```
shenglol, line 147: Could you clarify the reasoning behind blocking these expressions?

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+the use of non-secret parameters as well, such as the "namespace" in the Kubernetes extension use case. There could be
+a constraint to allow only non-secure parameter usage in extension configurations, but that could end up encouraging the
+user to work around that by not using secure parameters for data that should be secured. Alternatively, the secret portion
+of the extension configuration could be split into from non-secret data. The secret portion would adopt the design proposed
+in this REP and the non-secret portion could be kept as-is, but there would need to be a way for the extension to
+provide metadata about the properties of the configuration schema.
+
+The move of extension configuration definitions to the parent deployment properties along with the language expression 
+constraints it imposes means that extension resources that require a configuration will no longer be deployable
+in a root deployment. There is no mechanism at the root level to provide this data. Client side changes would need to
```
jeskew, line 273: If the `extensions` property is accepted on a nested deployment resource, that means it is part of the deployments PUT contract. Is the blocker here just adding new options to the CLI?

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+param clusterName string
+param namespace string = 'default'
+
+// duplicates a resource definition, but leaves the extension config definition site unchanged
+resource aksCluster 'Microsoft.ContainerService/managedClusters@2024-02-01' existing = {
+  name: clusterName
+}
+
+extension kubernetes with {
+  kubeConfig: aksCluster.listClusterAdminCredential().kubeconfig[0].value
```
jeskew, line 333: Shouldn't this be blocked? Line 145 says that references to deployment parameters are prohibited in extension configurations, but the `aksCluster.listClusterAdminCredential()` will be compiled to `[(listClusterAdminCredential(resourceId('Microsoft.ContainerService/managedClusters', parameters('clusterName')))...]`

active/0004-extensibility-support-for-deployment-stacks.md
shenglol: IMHO a temporary workaround could be to continue supporting the use of parameters to specify extension configurations within the module itself for the time being. We can decide later whether to introduce new CLI options or solely rely on .bicepdeploy files for extension configurations.

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+                  "namespace": {
+                     "type": "string",
+                     "value": "default"
+                  },
+                  "kubeConfig": {
+                     "type": "directive",
+                     "evaluation": [
+                        {
+                           "type": "ArmApiCall",
+                           "resourceId": "/subscriptions/.../resourceGroups/.../Microsoft.ContainerService/managedClusters/...",
```
jeskew, line 180: If the resource ID needs to be a compile-time constant, will this feature be usable in multi-cloud deployments? (Maybe it would be usable with Ev2 substitutions, but those aren't available to 3P users.)

active/0004-extensibility-support-for-deployment-stacks.md
shenglol: Although that creates some challenge as Kyle highlighted below in the alternatives section. A less ideal workaround is to ask customers to always declare a module to provide extension configs.

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+// The ARM aksCluster resource can supply kubeConfigs for the AKS Kubernetes use case.
+resource aksCluster '...' = {
+  // ...
+}
+
+module extResources 'extResources.bicep' = {
+  name: 'extResourcesDeploymentName'
+  extensionConfigs: {
+    k8s: {
+      kubeConfig: aksCluster.listClusterAdminCredential().kubeconfig[0].value
```
anthony-c-martin, line 111: Is this the "recommended" way to authenticate for kubernetes? Are there any other options available (e.g. Entra auth)?

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+```bicep
+// The ARM aksCluster resource can supply kubeConfigs for the AKS Kubernetes use case.
+resource aksCluster '...' = {
+  // ...
+}
+
+module extResources 'extResources.bicep' = {
+  name: 'extResourcesDeploymentName'
+  extensionConfigs: {
+    k8s: {
```
anthony-c-martin, line 110: I feel like we should be able to provide a simpler user experience for AKS, which feels like probably the 95% use case on ARM. It may also make life easier for Stacks in terms of being able to source a relationship between "scope" (AKS resource) and "child" (Kubernetes resource).

E.g. coming up with a concept of an "adapter" in Bicep from AKS resource -> Kubernetes resource :
```bicep
module extResources 'extResources.bicep' = {
  name: 'extResourcesDeploymentName'
  extensionConfigs: {
    k8s: aks.asScope()
...
```

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+main.bicep - The root deployment
+```bicep
+// The ARM aksCluster resource can supply kubeConfigs for the AKS Kubernetes use case.
+resource aksCluster '...' = {
+  // ...
+}
+
+module extResources 'extResources.bicep' = {
+  name: 'extResourcesDeploymentName'
+  extensionConfigs: {
```
anthony-c-martin, line 109: [nit] Would be nice to come up with a simpler name for this - what about just `extensions`?

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+  }
+  params: {
+    // ...
+  }
+}
+```
+
+extResources.bicep - Child deployment with extension resources
+```bicep
+extension kubernetes as k8s // optional alias
```
anthony-c-martin, line 123: Could you cover what the UX looks like for child -> grandchild module? 

E.g. it would be nice to avoid having to write the following if we're able to detect and automatically flow down the boilerplate:

```bicep
extension kubernetes

module foo 'foo.bicep' = {
  extensionConfigs: {
    kubernetes: kubernetes
  }
}
```

/cc @rynowak

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+```
+
+With the deployments service returning a list of all extensions (across all nested deployments) with their configuration
+directives and returning all extension resources deployed in the template (across all nested deployments), the stacks 
+service can process each extension resource and use the "deploymentId" and "extensionAlias" properties as a composite 
+key to the data in the extensions array.
+
+## Drawbacks
+
+The biggest drawback of this approach is that extension configuration data cannot be sourced from deployment parameters.
```
anthony-c-martin, line 261: I'm on the fence about whether we should completely block this. Have you considered supporting it, but with a "degraded" experience? I think that could be understandable from a user standpoint.

For example, you can still bring your own credentials for managing an on-prem Kubernetes cluster through Stacks, but it's not going to be resilient to rotations. As long as the user has some mechanism to rotate the credentials, I think it's understandable that the Stack won't be able to delete the resource.

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+This design decision is driven solely to support the stacks user experience not needing to prompt for the parameter
```
anthony-c-martin, line 262: Related to the above question - could there be any need to parameterize non-secure configuration? E.g. being able to parameterize the k8s namespace. Would we be blocking that with this approach?

kalbert312:
Follow up: look into possible permission issues with ARM API call directives. Can we do a first party call?

active/0004-extensibility-support-for-deployment-stacks.md
jeskew: It's also unclear how this would be derived within a Bicep compilation. Bicep never knows metadata like the subscription ID or RG name about the target scope -- Bicep emits ARM expressions like `subscription().subscriptionId` instead.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: As it's spelled out as of now, yes. I outline that a little bit below and in unresolved questions. I haven't had a chance to look into how the configuration object schema metadata is loaded or what its source is yet. If it could be modified to outline if a property is a secret/secure value or non-secret, that could potentially be used on the engine side to enable providing a resolved configuration back through the GET response. Namespace could be sourced from a non-secure parameter, and the value used for the deployment parameter can be persisted inside the configuration object for consumption by stacks.

The input extensions property to the deployment would need to be able to accept language expressions that would be evaluated at template evaluation time. There would need to be a way to throw an error if a secure value is used at template validation time.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: We would be able to support parameterization of configuration through ARM resources we can query live, key vault secret, etc, but not raw value secrets because we can't persist those. Stacks will need to be robust to failures when retrieving these secrets and provide clear errors regardless of how the constraints on secret sources are set up.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: Yeah the main blocker is a mechanism on frontends to provide extension configs but only in formats that can be compiled to the format expected in the PUT contract. It's like a parameters file but without accepting dynamic inputs (inline parameters in the CLI).

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: In this case, the engine could evaluate this parameter because it's not a secure parameter and persist the value inside the `ArmApiCall` directive in the `resourceId` shown in the ARM template example. It's an alternative that would allow parameter use in this case because it only depends on the name of the resource and scope detail.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: I like the idea for this specific resource. The Bicep compiler should know about the extension version so it seems like it would be able to support whatever configuration each version needs. It seems we'd also need to be able to parameterize the function with the namespace.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: I know you mentioned this to me some time ago. I took this from the official quick starts. I'm not super familiar with kubernetes or auth mechanisms available yet. It might be possible to provide other inputs for auth and funnel that to that the V2 extension implementation. I'll look into this and see if it affects the design at all.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: Good point. This will need to be implemented because passing the data as deployment parameters   would be non-sensical.

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+
+## Detailed design
+
+> The core of the REP. Elaborate on the design and implementation with sufficient detail for someone familiar with Bicep
+to grasp. Provide detailed examples to illustrate how the feature is used and its implications on user experience.
+
+The key design change is to move where extension configurations are defined in Bicep and ARM templates. Currently,
+the extension configurations are defined within the deployment of the extension resources being deployed. This REP's 
+proposal is to move these configuration definitions to a dedicated deployment input property that is separate from 
+deployment parameters. See the examples section for Bicep and ARM template examples.
```
rynowak, line 53: This is a huge improvement. Extension config should be first-class and part of the deployment resource rather than the template. We implemented this in the Radius fork of the DE and have been living with it happily for some time.

active/0004-extensibility-support-for-deployment-stacks.md
rynowak: Yes please. Extension config will be much easier to use if it automatically flows across module boundaries. 

We implemented this in the Radius DE fork and have been living with it happily for a while.

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+                     "type": "JSONPath",
+                     "path": "kubeconfig[0].value" // decompose?
+                  }
+               ]
+            }
+         }
+      }
+   ]
+}
+```
```
rynowak, line 252: I like the example with creating an AKS cluster and then passing in the kubecontext to a child module.

What's the "right" way to do this for a stack when the extension config is passed at the top level? 

- Imagine my AKS cluster already exists, do we need additional support for the CLI to work with extension config?
- Imagine it's not an AKS cluster 😆 should I store the config in keyvault?


Your

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,405 @@
...
+
+### Key vault secret expressions in extension configurations and other expression limitations driven by stacks
+
+Whether key vault secret access should be supported in extension configurations is a tricky subject. Users will likely
+want this feature, and it would work fine with a standalone deployment that isn't tethered to a stack because the secret
+is only needed at the time of deployment. The problem from the stacks perspective with introducing this feature is 
+that accessing a user's key vault in the future for the deletion of stack resources introduces a failure point. What if
+the user did not properly maintain this secret and it's now invalid? This type of issue would block the deletion or 
+update of a stack when resources that no longer managed are to be deleted and potentially result in unexpected stack
+states. The errors would likely manifest as incidents for the stacks team to investigate.
```
rynowak, line 382: This feels like a problem that can't be solved by stacks. If the data is gone, it's gone. For example in your AKS scenario - I could go out of band and delete the cluster 👿 

What's the right behavior when this happens?

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: 1. One way might be to extend the bicepparam/json file with an extensions section. It would allow referencing ARM resources but prohibit other sources (inlining, loadFileContent, env vars, etc) for secure portion of the configuration schema. Other bits like `namespace` could be expressed like parameters are.

2. Yes, this will be supported. That specific example is not outlined here yet. I think the general constraint to start with is that expressions are restricted to ARM resource accesses expressions. I detailed this some more in recent commits to the Bicep example.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: I've updated this to accept language expressions. Before the nested deployment is submitted, the extensionConfigs language expressions will be evaluated against the parent deployment template. The GET response will return the evaluated copy of this block. To support passing extension configurations down to nested deployments deeper than 1 level, I'm looking into an "inherit" approach where the template engine will substitute an "inherit" directive in the nested deployment properties with the evaluated copy provided to the current deployment before it's PUT.

kalbert312:
> > Follow up: look into possible permission issues with ARM API call directives. Can we do a first party call?
> 
> I think 1P calls have their own drawbacks to consider. Let's say you deploy an AKS cluster but your permissions on that cluster are subsequently revoked. Should you be able to delete it just because you used a Stacks API and thus bypassed the authZ check?

Yeah that shouldn't be allowed. We'll just need to handle these errors and present them to the user.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: I updated the design to include language expression parsing.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: I expanded on this in recent commits. There will need to be an "inherit" mechanism for "extensionConfigs" in the ARM template level. I believe the substitution should be able to be done within the DeploymentResourceJob for deployment resource types as that has access to the deployment entity + child deployment properties.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: This needs some more thought and is dependent on what various auth workflows are supported by AKS.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: Added design proposals for bicepparam/JSON parameter file changes.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: Adapted the design to accept language expressions. It should have been there from the start.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: Support for BYO credentials will be there, but only in ways where a fetch is done. No raw values. Current design has details on that.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: I updated the design to propose changes to the parameters file to support extension configs.

active/0004-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,665 @@
...
+  // ...
+}
+
+module extResources 'extResources.bicep' = {
+  name: 'extResourcesDeploymentName'
+  extensions: {
+    k8s: {
+     ✅ kubeConfig: aksCluster.listClusterAdminCredential().kubeconfig[0].value
+     ❌ kubeConfig: loadFileContent('kubeconfig.txt')
+     ❌ kubeConfig: secureParam
```
shenglol, line 176: I believe the two use cases mentioned can still be valuable for non-stack deployments, and customers might still want to use them. Rather than disallowing them, could we introduce a linter rule to indicate that they won't function with stacks?

active/0004-extensibility-support-for-deployment-stacks.md
shenglol: I like `extensions` as well for its simlicity and thought about using it, but that would introduce a conflict with the deployment output `extensions` property.

kalbert312: Done.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: Added linter rule design.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,713 @@
...
+                }
+              }
+            }
+          },
+          "template": {
+            "extensions": {
+              "k8s": {
+                "extension": "Kubernetes",
+                "version": "1.0.0"
+                // "config" is removed, it's expected to be provided by properties of the deployment
```
shenglol, line 343: We may need to keep the config type information to enable Bicep type validation for remote modules with extension imports, as the remote module artifact is essentially an ARM template.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,713 @@
...
+                },
+                "kubeConfig": {
+                  "type": "armApiCall",
+                  "apiReference": {
+                    "method": "GET",
+                    "resourceId": "[ /subscriptions/.../resourceGroups/.../Microsoft.ContainerService/managedClusters/... ]",
+                    "apiVersion": "2024-02-01",
+                    "path": "/listClusterAdminCredentials",
+                    "query": "[ ?a=1&b=2 ]",
+                    "valueJsonPath": "[ kubeconfig[0].value ]"
```
shenglol, line 332: This seems like an implementation detail to me that could be handled on the back end. Converting an expression to this object structure would require the same amount of work, regardless of whether it’s done in the Bicep compiler or the deployment engine. What’s the advantage of this approach compared to simply emitting a template expression as the value and processing the expressions within the deployment engine?

kalbert312: I'm undecided on this. I think either option would work and it depends on the complexity of each implementation. I will be digging into implementation details soon in my branch.

kalbert312: One reason for the API reference approach is it's similar to how key vault references are expressed so these API references could resolved in the same code path as key vault references are now. For the value language expression, I'm uncertain currently where this translation would take place as I need to spend time on that area of the code.

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: You're right, I left this detail out in this proposal. I meant the value-based portion. I'll update this to include the metadata needed for validation.

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: We may be able to support both, but we'd need the API reference variant at the root level (in params file) to source the kubeconfig from an AKS cluster if we're not using Azure based auth.

kalbert312: I spent more time tracing code and thinking about stacks vs non-stacks. We'll have to go with general language expressions.

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: Fixed.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,711 @@
...
+{
+  "rules": {
+    "stacks-compatibility": {
+      "level": "warning",
+      // The default level across all files. Up to user to configure as an error or to turn off.
+      // Alternative to file overrides is to stick with in-file/line overrides.
+      "fileOverrides": {
+        "*.stack.bicep": "error",
+        "<glob>": "error"
+      }
```
kalbert312, line 233: Bicep discussions 10/14: 
- Make it consistent with existing naming schemes.
- Break into smaller rules for maintainability.
- Default to informational.
- Hold off on file overrides for now. Collect feedback once rule is released.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,711 @@
...
+                "namespace": {
+                  "value": "[parameters('namespace')]"
+                },
+                "kubeConfig": {
+                  "value": "[listClusterAdminCredential(resourceId('Microsoft.ContainerService/managedClusters', parameters('clusterName')), '2024-02-01').kubeconfigs[0].value]",
+                  // When deployed as a stack, the value expression will be decomposed on the backend to acceptable 
+                  // formats returned by the deployment GET response. More on this in a further section.
+                }
+              }
+            }
```
shenglol, line 331: My appologies, I thought I left a comment earlier, but maybe I forgot to submit it 😑...

I'm considering whether we could simplify this by avoiding type as a discriminator.

For example:

```jsonc
// Explicitly specify the config object
{
  "k8s": {
    "namespace": {
      "value": "..."
    },
    "kubeConfig": {
      "value": "..."
    }
  }
}
```

```jsonc
// Inherit from the parent template
{
  "k8s": "k8s" // Use the extension symbol name from the parent template
}
```
The JToken type of "k8s" (`object` vs. `string`) should be enough to indicate whether the extension config object is explicitly defined or inherited.

Alternatively, instead of using symbolic name for inheritance, we could introduce an extensionConfigs function to access extension configurations directly:

```jsonc
{
  "k8s": "[extensionConfigs('k8s')]"
}
```
This approach gives users finer control over individual configuration items:

```jsonc
{
  "k8s": {
    "kubeConfig": {
      "value": "[extensionConfigs('k8s').kubeConfig]"
    },
    "namespace": {
      "value": "anotherNamespace"
    }
  }
}
```

What do you think?

kalbert312: I think an `extensionsConfig` function would be useful and necessary, otherwise there's no way to combine a single kube config with multiple namespaces for example. This highlights a gap in this current design that should probably be addressed sooner than later. We don't have a way to inherit and override a subset of values yet, or conditionally inherit an extension config as a whole.

I'm not really a fan of reducing the type of the root object of an extension's config to JToken (string & object). My preference there would be to keep things as strongly typed as possible for known schemas such as the reference models.

I see object spreads codegen to something like this:
```json5
"value": "[shallowMerge(createArray(variables('objA'), createObject('prop1', 'val')))]"
```

With the current implementation doing validation at preprocessing time to check secure data usage and support API reference conversion, it feels like something similar to the object spread syntax above complicates the preprocessing logic. It seems we'd need to optimistically evaluate a statement like this to validate the overridden parts. Runtime references would prevent optimistic evaluation so analyzing the function stack becomes more complex.

What do you think of the below model?

```json5
{
  "k8s": {
    "type": "inherit",
    "alias": "[if(..., 'k8s1', 'k8s2')]",
    "value": { // for inherits, this would override portions, preprocessing would only check this
      "namespace": {
        "value": "otherNamespace"
      },
      "otherProp": {
        "value": "[extensionConfigs('k8s1').otherProp]"
      }
    }
  }
}
```

This does feel like it would complicate the Bicep codegen though to produce this alternative output for object spreads and the alias expression...

kalbert312: Thinking about the model I suggested some more.. I don't think we'd support spreading reference sourced objects onto an extension config because we won't know the properties until the resource job runs. I guess if we wanted that we could possibly rerun the preprocessing like validation in resource job after references are resolvable. This syntax also doesn't support spreading an object sourced from a parameter, which depending on how complicated extension configurations could get, might be something that is desirable to do. If we wanted to support that we'd need another way to express that operation.

jeskew: @kalbert312 Would it help if you were able to leverage the folding evaluator we recently added to support what-if? Simplifying expressions that aren't fully evaluable might allow Stacks to support more complex expressions.

kalbert312: @jeskew Possibly. I haven't had a chance to familiarize myself with it yet.

kalbert312: Bicep discussions 11/4:
- Leave as is for now as a stopgap to a purely LE approach. The challenge is that we must know the root object properties that will be embedded in a LE without reducing the property value LEs so they can be validated.
- Add support for inheritance with overrides.
- Remove/rename confusing "type" property on the declaration.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,857 @@
...
+resource aksCluster '...' = {
+  // ...
+}
+
+module extResources 'extResources.bicep' = {
+  name: 'extResourcesDeploymentName'
+  extensionConfigs: {
+    k8s: {
+      auth: {
+        kubeConfig: aksCluster.listClusterAdminCredential().kubeconfigs[0].value
```
shenglol, line 115: If I understand correctly, the essence of @anthony-c-martin suggestion (please correct me if I'm wrong) is to use a discriminator to annotate the credential source, e.g., 

```
{
  k8s: {
    // KeyVault
    kubeConfig: {
       source: 'KeyVaultSecret'
       keyVault: myKeyVault
       secret: 'mySecret' 
    }
    // List secret
    kubeConfig: {
       source: 'ResourceSecret'
       resource: myAksCluster
       credentialType: 'Admin' // or 'User'
    }
    // User provided - not supported by Stacks
    kubeConfig: {
       type: 'UserProvided'
       value: '...' // or any explicit value, can even be aksCluster.listClusterAdminCredential().kubeconfigs[0].value
    }
}
```

This avoids the need to analyze expressions, as Stacks can use the credential source type discriminator to determine whether it is safe to save the credential reference. This approach differs from the current expression-based solution.

Have we considered this? If we move forward with the expression-based solution, I don’t see much benefit in maintaining a dedicated `auth` section.

kalbert312: Meeting 12/2: 
1. Change design to use this pattern for auth properties.
2. We can probably eliminate the "auth" wrapper object.
3. Check inheritance patterns against new design.
4. Future: investigate the extension providing ARM deployment engine the way to fetch credentials. For now, this will be hard coded in the deployment engine.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,821 @@
...
+        resource: aksCluster
+        credentialType: 'Admin' // or user
+      }
+      // a kubeConfig sourced from a user expression (NOT allowed for stack deployments):
+      kubeConfig: { //
+        source: 'Custom'
+        value: aksCluster.listClusterAdminCredential().kubeconfigs[0].value 
+      } 
+      // a kubeConfig sourced from a Key Vault. Note: this will be compiled to the reference object format (source = 'KeyVault')
+      kubeConfig: kv.getSecret('myKubeConfig') // alternatively: { source: 'KeyVault' value: kv.getSecret(...) }
```
shenglol, line 129: From a user perspective, I like the `kv.getSecret('myKubeConfig')` syntax because it’s less verbose and aligns with how KV secrets are used in parameters. However, it introduces a few challenges:

- The user experience becomes inconsistent with the other two source types.
- Bicep doesn’t support union types where members can be either strings or objects with discriminators. For example, the following type definition results in an error:

image

The alternative approach `{ source: 'KeyVault', value: kv.getSecret('...') }` resolves the issues above, but it comes with added verbosity. Another drawback is that even with the `source: 'KeyVault'` discriminator, expression-level validation for value would still be necessary, as users could directly modify the generated ARM templates.

Given that expression validation cannot be avoided, a possible solution could involve leveraging credential source annotations to simplify expression analysis in both Bicep and the deployment engine.

## Source: Resource secret

Using the kubeConfig sourced from an AKS cluster as an example, instead of requiring users to explicitly specify an object discriminator, the extension provider could annotate the ResourceSecret source with metadata in the type manifest, such as:

- Allowed Resource Type: `'Microsoft.ContainerService/managedClusters'`
- Allowed List Operations: `listClusterAdminCredential` and `listClusterUserCredential` (but perhaps excluding `listClusterMonitoringUserCredential`?)
- 
This approach would allow Bicep to perform semantic validation based on the provided type metadata and enable users to write concise expressions for defining the kubeConfig value. E.g.,

```
{
  // Valid
  kubeConfig: aks.listClusterAdminCredential().kubeConfigs[0].value
  kubeConfig: aks.listClusterUserCredential().kubeConfigs[0].value

  // Invalid - Not an allowed list operation
  kubeConfig: aks.listMonitoringUserCredential().kubeConfigs[0].value

  // Invalid - Security risk sending a storage account key to the Kubernetes extension
  kubeConfig: storageAccount.listKeys().keys[0].value

  // Invalid - Any invalid expressions within the syntax tree are blocked
  kubeConfig: someCondition
    ? storageAccount.listKeys().keys[0].value
    : 'someValue'
}
```

The generated ARM template for the valid case would look like this:

```
{
  "kubeConfig": {
    "value": "[listClusterMonitoringUserCredential(resourceId('Microsoft.ContainerService/managedClusters', 'myAKSCluster'), '2024-08-01').kubeconfigs[0].value]"
  }
}
```

We would need to replicate the same semantic analysis performed on the Bicep client side within the deployment engine to block invalid list operation calls. This could be achieved using hard-coded extension facts.

The process should be straightforward, as we can walk through the expression syntax tree, scanning for list function calls to verify their validity. As long as we confirm that the value represents a valid AKS cluster list* property access expression, it can be safely stored for Stack.

## Source: Key Vault Secret

Specify a Key Vault reference would be very simple:

{
  kubeConfig: kv.getSecret('myKubeConfig')
}

The generated ARM template would contain:

```
{
  "kubeConfig": {
    "reference": {
        "keyVault": {
          "id": "/subscriptions/<subscription-id>/resourceGroups/<rg-name>/providers/Microsoft.KeyVault/vaults/<vault-name>"
        },
        "secretName": "myKubeConfig"
      }
  }
```

The handling of this in the deployment engine is trivial.

## Source: Custom

Any value that does not adhere to the format of a list operation or a KeyVault reference, or includes overly complex expressions, can be classified as a custom value.

- Bicep Behavior: Bicep can emit `information` (or `warning`) diagnostics to inform users that such values will not be supported for Stack deployments.
- Deployment Engine Behavior: During Stack deployments, the deployment engine can validate these values and throw a runtime error if they are encountered.

```
{
  // Explicit string value. Treated as a custom value.
  kubeConfig: 'someExplicitValue'

  // Reference to a parameter. Treated as a custom value.
  kubeConfig: someParameter

  // Reference to a variable. Treated as a custom value.
  kubeConfig: someVariable

  // Conditional expression with a valid list operation.
  // Supporting this would require extensive expression analysis, which is likely unnecessary.
  // Users should use conditional modules for such scenarios instead.
  kubeConfig: someVariable
    ? aks.listClusterAdminCredential().kubeConfig[0].value
    : 'someValueWhenFalse'

  // Conditional expression switching between two AKS resources.
  // Complex expression that is not recommended due to analysis overhead.
  kubeConfig: (someCondition ? aks1 : aks2).listClusterUserCredential().kubeConfig[0].value

  // Overly complex indexing logic with list operations.
  // Why would someone do this?
  kubeConfig: aks.listClusterUserCredential()[someParameter][someSecureParameter]
}
```

The key here is to streamline maintenance by minimizing expression analysis. This approach avoids partial expression reduction and sidesteps obscure edge cases. Here's a straightforward processing flow for the deployment engine:

1. Security Check: Determine if the extension configuration item is secure using extension facts.

  - If not secure, it can be safely saved for Stack.

2. KeyVault Reference Validation: If the item is secure, check whether it is a KeyVault reference.

  - If it is a KeyVault reference, save it for Stack.

3. List Operation Validation: If not a KeyVault reference, scan for invalid list operations based on extension facts.

  - If it matches a valid list* property access expression and it does not contain any weird parameter / variable reference or condition expressions, save it for Stack.

4. Custom Value Handling: For any other cases that do not fit the above criteria:

  - Treat the value as a custom configuration.
  - Return an error for Stack deployments.

I believe the above solution offers several key benefits. It provides a straightforward and consistent approach for users to specify configuration items, minimizing confusion. Furthermore, it simplifies the complexity of extension facts - rather than storing detailed instructions for interpreting configuration items to retrieve credentials, it only requires tracking the resource type and a list of valid list operations. This approach is also flexible enough to be generalized for other extensions in the future.

kalbert312: About `kv.getSecret()`: this currently does compile to a key vault reference in the ARM template because it's only usable for secure nested deployment parameters so we wouldn't be unioning a string here. I was wondering if we should try to make it consistent to the nested deployment parameter style but I do agree it produces an inconsistency with the other sources. There's a couple choices on the Bicep side here:
1. Go full object literal for references, user must write out the object by hand.
2. Provide Bicep-only functions that compile to the reference format. This allow "Custom" expressions without the object wrapper and would require a special function for the "Resource" source (ex `resourceSecret(aks, { credentialType: '' }`) although this would more or equally be difficult to implement compared to the object literal due to the introduction of properties determined by the resource passed in). Maybe this can be worked around with your "resource secret" suggestion.

# Source: Resource secret
I'm wondering what the user experience will be with this. 

How will we guide the user to write an expression based on dynamic resource type(s) and with it, dynamic list* calls? I suppose there can be a couple diagnostics here, 1 for the expectation of a resource list* call of a set of resource types, and then a diagnostic for the list* part of the expression that lists out the acceptable ones. That paired with some examples in the documentation is probably enough.

We lose the extension guided value resolution with the object literal format and it off loads that logic back to the user. For example, take the AKS Admin vs User credentials case. With the literal format, the user could declare a strategy instead of a credentialType where the extension decides the right credential to use depending on information supplied to it, either from the template, deployment engine (submitting user for instance), or the configuration of the resource (the resource inputs, and any alteration some deployment script may make to it). I'm not sure if this specific example is an actual use case, but I think having the layer of `(deployment, template, resource) => credentialInstructions` in the extension would be beneficial to have. Do we want to give this up?

# Source: Key Vault Secret
I think we should go with this because this is the same format for deployment parameters.

# Source: Custom
Seems viable. We can check the inner most function. If it's a list* and there's no 'if's along the way we assume the resource source, otherwise custom.

shenglol: > About kv.getSecret(): this currently does compile to a key vault reference in the ARM template because it's only usable for secure nested deployment parameters so we wouldn't be unioning a string here

It compiles into an object in the ARM template; however, the return type of kv.getSecret() in Bicep is string. When specifying kubeConfig in Bicep, we must choose either object format or string expressions, but mixing both will not be supported.

shenglol: > I suppose there can be a couple diagnostics here, 1 for the expectation of a resource list* call of a set of resource types, and then a diagnostic for the list* part of the expression that lists out the acceptable ones. That paired with some examples in the documentation is probably enough.

That’s along the lines of what I was thinking. We could also enhance the experience by incorporating type-guided completions.

> We lose the extension guided value resolution with the object literal format and it off loads that logic back to the user.

Agreed. It does compromise flexibility, which is something I’d prefer not to lose. Providing users with a literal `Admin | User` credential type also feels like a much better experience overall.

I wonder if there’s a solution that could combine the strengths of both approaches, offering a concise and user-friendly experience while preserving flexibility...

shenglol: I have another idea! It still uses an object discriminator but avoids the caveats of union types:

```
{
  k8s: {
    // Descriminator for clusterType rather than kubeConfig
    clusterType: 'Managed | Custom'

    // clusterType == 'Managed'
    // Extension-guided resolution is applied.
    managedCluster: {
      id: aks.Id
      credentialType: 'Admin' | 'User'.
    }
    
    // clusterType == 'Custom'
    // KV reference - supported by Stacks.
    kubeConfig: kv.getSecret('kubeConfig')

    // Non-KV reference - not supported by Stacks.
    kubeConfig: someParam // Example of unsupported value
    kubeConfig: whatever  // Example of unsupported value
  }
}
```

What do you think?

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,821 @@
...
+
+The stacks service already sends a stack resource ID HTTP request header to the deployments service. This will be used
+to enable additional validation in the Deployment engine to enforce these expression rules. It will also enable the
+return of extension configurations in the Deployment GET response for API versions that support it. This header is only
+needed for the initial deployment PUT, as its value will be saved on the deployment metadata and this will be used to
+drive the API behavior. For deployments that are not initiated by the stack service, the configuration portion of the
+extension details will not be returned, as no additional validation on language expressions is done, which means we
+cannot guarantee compliance to the ARM RP RPC rule that a resource GET cannot return secret data.
+
+#### Stacks mode for Bicep files
```
majastrz, line 214: Should we consider alternative mechanisms for declaring that the file represents a stack? Should we introduce something in the .bicep files to let users indicate it?

kalbert312: I ended up deciding to stick to linter rules for this validation, thus the designation is in the bicepconfig.

My very first idea was to include a directive line in the file, something like `compatibility stacks`. These `compatibility` directives would only apply limitations, so it would be compatible with a deployment flow without them. It didn't seem necessary to introduce this given bicepconfig can scope the linting rules to a set of files and within the files, the linter rule can be manually ignored. Having linting rule(s) that default to informational will enable the user to see the considerations but also ignore them or enforce them.

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: > I have another idea! It still uses an object discriminator but avoids the caveats of union types:
> 
> ```bicep
> {
>   k8s: {
>     // Descriminator for clusterType rather than kubeConfig
>     clusterType: 'Managed | Custom'
> 
>     // clusterType == 'Managed'
>     // Extension-guided resolution is applied.
>     managedClusterId: aks.id
>     credentialType: 'User' | 'Admin' // Optional. Default is 'User'.
>     
>     // clusterType == 'Custom'
>     // KV reference - supported by Stacks.
>     kubeConfig: kv.getSecret('kubeConfig')
> 
>     // Non-KV reference - not supported by Stacks.
>     kubeConfig: someParam
>     kubeConfig: whatever
>   }
> }
> ```
> 
> What do you think?
This seems like a good alternative. I'll work through the details of this see if there's any issues.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,959 @@
...
+
+main.bicepparam
+
+```bicep
+using 'main.bicep'
+
+param boolParam = true
+// ...
+
+extensionConfigs {
```
shenglol, line 320: An alternative approach would be to reuse the `extension` and `with` keywords for supplying extension configurations, using a separate statement for each extension:

```
extension k8s with {
  namespace: 'default'
  ...
}
```

shenglol: To clarify, this isn't a suggestion. I'm simply thinking out loud and offering an alternative.

kalbert312: I agree this is better for the params file. Will update.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: This has been implemented in the latest design. The extension config will now accept properties that will resolve to the list* call on the engine side.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,1005 @@
...
+            "resources": {
+              "extResources2": {
+                // nested deployment (3rd level)
+                "type": "Microsoft.Resources/deployments",
+                "apiVersion": "2022-09-01",
+                "name": "extResourcesDeploymentName2",
+                "properties": {
+                  "extensionConfigs": {
+                    // whole inheritance:
+                    "kubernetes": "k8s", // ✅ extension name + version must match.
```
shenglol, line 547: Would we also allow `"kubernetes": "[extensionConfigs('k8s')]"`?

shenglol (approved):
Looks great!!

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: It can be added but there must be additional validation added in the visitor:
1. If we're at the extension alias path, then all LE branches must resolve to a root `extensionConfigs` usage otherwise the LE validation will be bypassed.
2. The `extensionConfigs` usages need to be validated to match the extension name and version of the target extension config so there is alignment.

shenglol: I'm wondering if we could delegate the validation of extension configurations entirely to the extension providers. In Bicep, we could still enforce type checking, but perhaps it would make sense to allow inheriting the entire extension without strictly matching the extension name and version.

From a user's perspective, there doesn’t seem to be a significant difference between:

```Bicep
extension kubernetes

module foo 'foo.bicep' = {
  extensionConfigs: {
    kubernetes: kubernetes
  }
}
```

and

```Bicep
extension kubernetes

module foo 'foo.bicep' = {
  extensionConfigs: {
    kubernetes: {
      kubeConfig: kubernetes.kubeConfig
      namespace: kubernetes.namespace
    }
  }
}
```

Users might question why only the first case is treated as special that requires a match on the extension name and version.

kalbert312: Yeah it is a little strange from that perspective. I included the name + version match to align the security characteristics of each property. Thinking about it however I think the frontline PUT validation will error out all cases where there could ever be a non-secure property overlayed to a secure property and vice-versa. A secure property for stacks can never have a literal value and a non-secure property can never be from a reference, so an error would always result.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,998 @@
...
+            "action": "listClusterAdminCredentials",
+            "query": "a=1&b=2",
+            "responseValuePath": "kubeconfigs[0].value"
+          }
+        }
+      }
+    }
+  ],
+  "outputResources": [ // "outputResources" is only populated when the deployment is successful. Stacks uses operations to peel through failed deployment resources.
+    { // NEW: Extensible resource
```
anthony-c-martin, line 700: In terms of swagger modeling, if we want to support polymorphism (e.g. "either an ARM resource, or an extension resource"), we may be asked to add a [discriminator property](https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/#discriminator) - e.g. either:
```json
{ "type": "arm", "id": "..." }
```
or:
```json
{ "type": "extension", ...other extension properties... }
```

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,998 @@
...
+      "extensionName": "Kubernetes",
+      "extensionVersion": "1.0.0",
```
anthony-c-martin, line 702: Is the repetition here necessary with extensionName & extensionVersion? Feels like if we have the configId, we should be able to look these up easily.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,998 @@
...
+    }
+  ]
+}
+```
+
+Operation GET for extensible resource:
+```diff
+{
+  ...
+  "type": "Microsoft.Resources/deployments",
```
anthony-c-martin, line 724: [nit] is this the correct resource type?

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,998 @@
...
+      clusterType: 'Managed'
+      managedClusterId: aksCluster.id
+      credentialType: 'Admin'  // 'Admin' | 'User'
+      
+      // END clusterType = 'Managed
+      
+      // BEGIN clusterType = 'Custom' w/ key vault
+      
+      clusterType: 'Custom'
+      kubeConfig: kv.getSecret(...)  // Only allowable expression type for secure properties for stack deployments. Other ARM resource sourced credentials must be supported by the extension via config properties.
```
anthony-c-martin, line 129: Should we differentiate "custom, sourced from keyvault" from "custom, supplied via expression"?

One thing that's caused us problems in the past is that the semantics of `getSecret()` w.r.t code generation are not clearly defined. This feels like a place we could improve that.

For example, we could define the type for this extension as (using typescript-like syntax):
```ts
type ManagedClusterType = {
  clusterType: 'Managed'
  managedClusterId: string,
  credentialType: 'Admin' | 'User'
};

type KeyVaultManagedClusterType = {
  // the name "clusterType" probably would need changing if we took this approach
  clusterType: 'KeyVaultManaged'
  kubeConfigReference: {
    keyVaultResourceId: string,
    secretName: string,
    secretVersion?: string,
  }
};

type UserProvidedClusterType = {
  // the name "clusterType" probably would need changing if we took this approach
  clusterType: 'UserProvided'
  kubeConfig: string,
};

type KubernetesExtensionConfig = {
    namespace: string
} & (ManagedClusterType | KeyVaultManagedClusterType | UserProvidedClusterType);
```

I think it would be worth going through the exercise of defining the types explicitly if we haven't already, to reduce the "magic" that happens with the `getSecret()` function.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,998 @@
...
+
+##### Limitations
+
+Because secure property values must be provided in reference format, there are several limitations on allowable Bicep
+expressions inside of extension configurations for any deployment:
+
+1. No object spreads within an expression scope that is not represented as a language expression in the ARM template.
+  This includes the top-level extension configuration object and secure property reference objects.
+1. Extension configuration inheritance is limited to whole inheritance or explicit piecemeal inheritance (each property
+  must be explicitly written out to inherit from some value).
```
anthony-c-martin, line 304: It would be good to revisit this to see if there's anything we can do to remove these limitations. Similar limitations in e.g. parameters adds complexity to code generation, and in the conceptual model that users have of the language.

active/0006-extensibility-support-for-deployment-stacks.md
```diff
@@ -0,0 +1,998 @@
...
+consuming deployment for the nested deployment.
+
+Here are some examples of passing a configuration down another level:
+
+```bicep
+extension kubernetes // this was provided a configuration through the parent deployment/parameters file
+
+// ✅ whole inheritance
+module foo 'foo.bicep' = {
+  extensionConfigs: {
```
anthony-c-martin, line 260: It would be good to discuss naming for `extensionConfigs` (both here, and the proposed function) - it makes sense, but feels quite verbose

active/0006-extensibility-support-for-deployment-stacks.md
anthony-c-martin: The other stuff I think this might help with is:
* Minimizing accidental mixing of resolved & unresolved secure values.
* Avoid the deployment engine having to be aware of the config schema
* Possibly simplify the analysis that Stacks would have to do to detect "user supplied" vs "key vault supplied"

shenglol: > the semantics of getSecret() w.r.t code generation are not clearly defined

@anthony-c-martin Do you mind clarifying this? I'm trying to understand the potential drawbacks of `getSecret()`. From a Bicep user's perspective, my understanding of the current semantics is that it's simply used for "retrieving a secret as a string from a KeyVault," which aligns with what the deployment engine does when processing the template. If the ARM template layer were removed between Bicep and the deployment engine, the current semantics would just seem appropriate. Also, from an extension provider's perspective, it might feel somewhat awkward to define the `KeyVaultManaged` type in the extension manifest, given that the KeyVault reference is not what they expect to receive when handling requests. In my view, the KeyVault reference type primarily makes sense as an intermediate "instruction" within ARM templates, and it's reasonable to abstract it away from both users and extension providers.

Regarding the distinction between user-supplied and KeyVault-supplied values, there doesn't appear to be a significant difference between checking the `value` and `reference` properties vs using `UserProvided` and `KeyVaultManaged` as discriminators. I think the former approach might even be simpler and more generic, as it avoids relying on specific property names / schema. Implementing the former should also be simpler since we could reuse the same code for handling KeyVault references in parameters.

Another concern with introducing `getSecretReference` is the inconsistency it creates between parameters and configurations. This could lead to confusion, as users would need to use `getSecretReference` in extension configurations but `getSecret` in parameters. If we decide to move forward with `getSecretReference`, it would make sense to apply it consistently across both parameters and configurations, which might be worth revisiting in the future.

shenglol: Had an offline discussion, and @anthony-c-martin highlighted these drawbacks of `getSecret`:

- Mixing raw secret values with `getSecret` can expose secrets. Retaining the concept of a secret as a reference at the control plane layer would enhance security.
- The implementation of `getSecret` in Bicep relies on hacky workarounds to support specific scenarios, like ternary expressions, but it cannot handle more complex expressions. There are quite a few Bicep issues related to this.

Overall, I agree that adding a new capability to allow users to "dereference a secret from a Key Vault without treating it as a raw secret value" would be valuable. However, I would still recommend addressing this issue for both parameters and extension configurations. Since designing a robust solution may take time, I wonder if this could be tackled in a future REP.

@anthony-c-martin @kalbert312 thoughts?

kalbert312: Decision was to defer this key vault reference issue to a separate project.

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: There was some back and forth on this early on. We ended up sticking with `extensionConfigs`. If I recall correctly, the reason was to keep it close to the request body format to avoid a layer of indirection, as resource definitions need to align with their PUT. @shenglol any thoughts? I know you were keen on keeping it `extensionConfigs` versus `extensions`.

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: Without the validation checks, we may likely be able to support this. It really depends on whether we decide to validate language expressions/value sources in the long term. The complex expressions would make the validation harder to carry out.

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: Config ID is nullable for extensions that do not require a configuration.

Even if it wasn't nullable, I think we should keep these in the equality check for configuration to be on the safe side. That way extension developers do not need to worry about collisions across extensions and we don't need to debug it if it were to ever occur. Will it happen? I'm not sure, but the extension footprint is already pretty generic, so it could happen. Examples I can think of:
 - Multiple extensions for something as big as Azure, and they all decide to use the management base URL as the config ID
 - Multiple extensions for an AKS cluster that all have their own abstraction layer over AKS.

I'd also expect the same extension but a differing version to return the same config ID for the same deployment target. The other data like resource identifiers will be tied to that version, so if we don't track versions used for particular resources and default to newest, there could be incompatibilities that creep up.

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: I originally wanted to add this. Right now it's ID vs identifiers, which is a little bit odd. I'm not sure if it's required a discriminator be in that format. Identifiers is also an object, so that complicates things. I'm up for adding it. @shenglol thoughts?

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: I removed the validation.

active/0006-extensibility-support-for-deployment-stacks.md
shenglol: I like the idea. Adding a discriminator will improve readability.

active/0006-extensibility-support-for-deployment-stacks.md
shenglol: One implication with the renaming the function `extensionConfigs` to `extensions` is that it will make more sense for the function to return not just the `config` object but also the extension `name` and `version`. This might be helpful if the user wants to output the extension name / version:

```Bicep
output k8sVersion string = k8s.version
```

```JSON
{
  "outputs": {
    "k8sVersion": "[extensions('k8s').version]"
  }
}
```

active/0006-extensibility-support-for-deployment-stacks.md
shenglol: On second thought, aren't the `extension` properties sufficient enough to distinguish between ARM and extensible resources?

anthony-c-martin: Discussed on a call - we should be fine to leave this as-is, and address it if it comes up in review.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: Azure auth is out of scope for this rep.

active/0006-extensibility-support-for-deployment-stacks.md
kalbert312: Removed.

active/0004-extensibility-support-for-deployment-stacks.md
kalbert312: For resources, a resource that is not found (204/404) will be considered already deleted. For the cluster deletion scenario that's more involved. We'll need a specific error code from the extension to detect a "control plane doesn't exist" case. In that specific case, from the stacks perspective, we could consider the resource deleted.

kalbert312: Essentially, it will be treated the same as a deleted resource group.

PR #8, opened 2024-08-28
Extensible Resource Deployment Observability Improvements

kalbert312:
Recreation of #3.

Includes changes from my feedback on that PR.

### DIFF
--- a/active/0003-extensible-resource-deployment-observability-improvements.md
+++ b/active/0003-extensible-resource-deployment-observability-improvements.md
@@ -218,6 +218,10 @@ Implementing this enhancement involves comprehensive modifications across variou
 
 Building on the previous discussion, if the deploymentProvider details are not included within `outputResources` for Azure resources, should we consistently list the `az` provider in the `deploymentProviders` for sake of consistentcy and clarity?
 
+### Should `outputResources` include `existing` resources as well?
+
+Currently, `outputResources` only contains created or updated resources. Should we expand the property to include `existing` resources?
+
 ## Out of scope
 
 ### Portal UI design
--- a/active/0003-extensible-resource-deployment-observability-improvements.md
+++ b/active/0003-extensible-resource-deployment-observability-improvements.md
@@ -30,22 +30,24 @@ Users currently encounter a major challenge when deploying extensible resources,
 
 To address this issue, the deployments and deployment operations APIs must be updated to include additional information, such as:
 
-- Resource symbolic name
-- Resource type
-- Deployment ID
+- Deployment resource ID
 - Bicep extension alias
 - Bicep extension name
 - Bicep extension version
+- Resource symbolic name
+- Resource type
+- Resource API version
+- Resource identifiers
 
 This additional data would enable users to identify a deployed extensible resource and map it to the corresponding resource definition in a Bicep file or an ARM template, providing users with comprehensive insights.
 
 ## Detailed design
 
-### `Microsoft.Resources/deployments` API changes for PUT and GET
+### API change for `GET /Microsoft.Resources/deployments/{deploymentId}`
 
 #### Adding `extensions`
 
-The `Microsoft.Resources/deployments` API will be updated to include an `extensions` array in both the PUT and GET response bodies. This modification is intended to incorporate information about extensions that are involved in executing the parent deployment and nested deployments to offer a more comprehensive deployment view.
+The `Microsoft.Resources/deployments` API will be updated to include an `extensions` array in the GET response body. This modification is intended to incorporate information about extensions that are involved in executing the parent deployment and nested deployments.
 
 ```diff
 {
@@ -78,7 +80,10 @@ The `Microsoft.Resources/deployments` API will be updated to include an `extensi
 +       "deploymentId": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Resources/deployments/my-deployment",
 +       "alias": "kubernetes",
 +       "name": "Kubernetes",
-+       "version": "1.27.8"
++       "version": "1.27.8",
++       "config": {
++         // Out of the scope of this REP, but it is required to enable Deployment Stacks integration.  
++       }
 +     }
     ]
   }
@@ -87,7 +92,7 @@ The `Microsoft.Resources/deployments` API will be updated to include an `extensi
 
 #### Updating the `outputResources` to include more identifiable information
 
-To improve the current structure where each `outputResources` object only has a resource `id`, it's essential to add more properties for identifying extensible resources, which lack resource IDs. These additions should include the symbolic name, the resource type, the ID of the deployment containing the resource, and the deployment provider alias.
+To improve the current structure where each `outputResources` object only has a resource `id`, it's essential to add more properties for identifying extensible resources, which lack resource IDs. These additions should include the resource symbolic name, resource type, resource API version, resource identifier, the resource ID of the deployment containing the resource, and the Bicep extension alias.
 
 ```diff
 {
@@ -101,21 +106,29 @@ To improve the current structure where each `outputResources` object only has a 
       },
       // Extensible resource
       {
-+       "symbolicName": "myService",
-+       "resourceType": "core/Service@v1",
 +       "deploymentId": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Resources/deployments/my-deployment",
 +       "extension": "kubernetes"
++       "symbolicName": "myService",
++       "resourceType": "core/Service",
++       "apiVersion": "v1",
++       "identifiers": {
++         "metadata": {
++           "namespace": "default",
++           "name": "myService",
++         },
++         "serverHostHash": "60fd32871cbe255d5793ce9e6ebf628beba25224cde7104e6a302f474f2f656e"
++       }
       }
     ]
   }
 }
 ```
 
-### `Microsoft.Resources/deployments/operations` API changes for GET and LIST
+### API changes for `GET /Microsoft.Resources/deployments/operations/{operationId}` and `GET /Microsoft.Resources/deployments/operations`
 
-#### Updating `targetResource` to include symbolic name and extension for extensible resources
+#### Updating `targetResource` to include extensible resource and extension information
 
-To align with modifications made to the `Microsoft.Resources/deployments API`, the response body now incorporates the addition of `symbolicName` and `extension`.
+To align with modifications made to the `Microsoft.Resources/deployments` API, the response body now incorporates the addition of `symbolicName`, `resourceType`, `apiVersion`, `identifiers`, and `extension`.
 
 ```diff
 {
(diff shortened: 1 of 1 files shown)


active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,282 @@
+---
+REP Number: 0003
+Author: shenglol (Shenglong Li)
```
shenglol, line 3: Feel free to update the author to yourself.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,311 @@
...
+    "outputResources": [
+      // Azure resource
+      {
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Storage/storageAccounts/my-storage-account"
+      },
+      // Extensible resource
+      {
++       "deploymentId": "/subscriptions/00000000-0000-0000-0000-000000000001/resourceGroups/my-resource-group/providers/Microsoft.Resources/deployments/my-deployment",
++       "extensionAlias": "k8s",
++       "extension": "Kubernetes",
```
shenglol, line 111: Customers should be able to use extensionAlias and deploymentId to look up the extension name. Was `extension` added solely for convenience?

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,311 @@
...
+> - The concept of "extensibility" has been a recurring theme in our public documentation. Adopting the term `extensions` aligns closely with this existing terminology, making it a more intuitive choice.
+> - 💪 "Bicep extension" resonates well with our established naming conventions for Bicep-related features.
+
+
+## Out of scope
+
+### Portal UI design
+
+The update to the Portal deployments blade is necessary to display the symbolic name and Bicep extension information for extensible resources. Since the portal team has ownership of the deployments blade, they hold the responsibility for both the UI design and the implementation of these changes. The addition of new properties — `extensions` in the deployments API and `extensions` in the deployment operations API — provides the Portal team with the essential data to programmatically differentiate between Azure resources and extensible resources. This capability is key to enabling the team to design a UI that effectively segregates and presents Azure resources and extensible resources.
```
anthony-c-martin, line 307: I think it's fine to consider this out of scope, as long as we're confident there's no ambiguity around whether the proposed API changes give them the right data to build the experience. Is it worth engaging to double check this (e.g. have the Portal team do a mock-up)?

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,311 @@
...
+```diff
+{
+  ...
+  "properties": {
+    "provisioningOperation": "Create",
+    "provisioningState": "Succeeded",
+    "targetResource": {
+      "resourceType": "core/Service",
++     "apiVersion": "v1",
++     "symbolicName": "myService",
```
anthony-c-martin, line 143: Is `symbolicName` going to be added to the response for ARM resources?

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,311 @@
...
+#### Updating the `outputResources` to include more identifiable information
+
+To improve the current structure where each `outputResources` object only has a resource `id`, it's essential to add more properties for identifying extensible resources, which lack resource IDs. These additions should include the resource symbolic name, resource type, resource API version, resource identifier, the resource ID of the deployment containing the resource, and the Bicep extension alias.
+
+```diff
+{
+  ...
+  "type": "Microsoft.Resources/deployments",
+  "properties": {
+    "outputResources": [
```
anthony-c-martin, line 102: Do we need any changes to the `dependencies` property? I wasn't sure whether it's currently supported for extensible resources.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,311 @@
...
+
+The `Microsoft.Resources/deployments` API will be updated to include an `extensions` array in the GET response body. This modification is intended to incorporate information about extensions that are involved in executing the parent deployment and nested deployments.
+
+```diff
+{
+  ...
+  "type": "Microsoft.Resources/deployments",
+  "properties": {
+    ...
+    "providers": [
```
anthony-c-martin, line 58: Off-topic, but I wonder if anyone actually uses this. Maybe we should consider cutting it to reduce confusion?

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,311 @@
...
+
+```
+VERBOSE: 00:00:00 - Extensible resource with symbolic name 'myService' of type 'core/Service@v1' has provisioning status: 'Succeeded'. The resource is provisioned by 'Kubernetes@1.27.8'.
+```
+
+## Drawbacks
+
+### Inconsistency between Azure resources and extensible resources
+
+The new properties are added exclusively for extensible resources, leaving Azure resources unchanged, which avoids introducing breaking changes to the API, SDK, and clients. However, this approach leads to an inconsistency in the schema between Azure resources and extensible resources. Such a disparity could potentially cause confusion, as users might expect a uniform schema across all resource kinds. Clear documentation might be required to manage user expectations and explain the rationale behind the differing schemas.
```
anthony-c-martin, line 175: We might need to consider what the behavior is for ARM resources that have been declared as pseudo-extensible reosurces. See [here](https://msazure.visualstudio.com/One/_git/AzureUX-Deployments?path=/src/Templates.Tests/TemplateSymbolicNameTests.cs&version=GBmaster&line=486&lineEnd=523&lineStartColumn=1&lineEndColumn=16&lineStyle=plain&_a=contents) for an example.

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,311 @@
...
+    "resourceType": "core/Service",
+    "apiVersion": "v1",
+    "identifiers": {
+      "metadata": {
+        "namespace": "default",
+        "name": "myService",
+      },
+      "serverHostHash": "60fd32871cbe255d5793ce9e6ebf628beba25224cde7104e6a302f474f2f656e"
+    },
++   "identifiersHash": "SHA-256 hash that can be used as a key"
```
kalbert312, line 232: Do not add to API contract. Stacks service will handle any primitive id needs.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: Follow up: Need to determine if we need dependencies for the first pass or if a brute-force retry will be sufficient.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: Follow up: delete `extension` property. The alias and deploymentId will be the key to the extensions array.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: I removed it from the code sample. I'm not sure why it is there since I inherited this REP. I'm going to leave anything pertaining to this property out of this REP so it focuses on just the extensions/outputResources/operation changes for extensions.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: It should provide enough data to at least list the resources in the deployment details pane. However, if a clickable link to the extension resource should be provided, I'm not so certain how this would be formulated. For Azure resources, the resource ID itself is likely enough. For the AKS example, it's unclear how to link a specific AKS cluster to each individual resource in a way the portal understands. Maybe this is something the extension config REP can address. One solution could be to include a list of referenced ARM resources in the configuration detail. The portal UI could find an AKS cluster resource in that list and if it exists, associate each resource scoped to the extension to the cluster.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: I'm not certain around the original points that led to that decision, but my guess is related to potential complications between non-symbolic and symbolic resource templates.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: I'm thinking the decision to not modify the Azure resource schema should be reevaluated. To me it is strange that particular case versus non-extensible case would produce different results in the API. 

I think the decision not to change the Azure resource schemas was driven by 2 points:
1. Prevent regression in downstream systems for non-extension deployments by preserving schema.
2. Complications from varying deployment feature status (symbolic resource names for example).

I can highlight this topic in the upcoming discussion meeting.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: I don't think we'll need changes for stacks extensibility support. I added a note about this in the unresolved questions. For deletion, we'd only want to know about container relationships (like that of an RG and a resource inside of it) and not data-only dependencies (such as using the output properties of some resource as input to another).

active/0003-extensible-resource-deployment-observability-improvements.md
```diff
@@ -0,0 +1,298 @@
+---
+REP Number: 0003
```
shenglol, line 2: May need to change this to `0008` to match the PR number.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: Followup: Create a backlog item for clickable links for ext resources.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: Decision was to investigate adding properties only under the feature flag and new API version using data that may already be available in the code context. Properties that may not be so simple to add may be omitted as necessary.

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: Decision is to figure out if symbolic name is feasible to add now. It may be omitted in the implementation.

kalbert312: I added this for at least the output resources in the new API changes. There's already metadata on the job for it so it should be easy to add to the operations as well.

shenglol (approved):

active/0003-extensible-resource-deployment-observability-improvements.md
kalbert312: Adding symbolic names to ARM resources for the operation API is deferred for now because sourcing the data is not so easy. Deployment resource reference does not have this data and would require significant refactors to supply it.

kalbert312:
I'm going to close in this in favor of including this design in https://github.com/Azure/bicep-reps/pull/6. Recent iterations of that change the fields added in this REP to handle various stacks deployment edge cases in tracing extensible resources.

PR #10, opened 2024-10-29
User-defined type validation decorators

jeskew:
Addresses https://github.com/Azure/bicep/issues/2922

### DIFF
--- a/active/0000-custom-validation-decorators.md
+++ b/active/0000-custom-validation-decorators.md
@@ -1,0 +1,242 @@
+---
+REP Number: <Fill me in with a four-digit number matching the pull request number; Update AFTER PR is approved and BEFORE is merged.>
+Author: jeskew (Jonathan Eskew)
+Start Date: 2024-10-29
+Feature Status: Public Preview
+Bicep Issue Number(s): 2922
+---
+
+<!-- Remove this comment and the prompts (in the form of blockquotes) for each section before submitting your PR -->
+
+# Custom validation decorators
+
+## Summary
+
+Template authors should be able to define custom decorators using Bicep's user-defined function syntax. As a first
+application of this affordance, template authors should be able to declare custom validation decorators that may be
+applied anywhere a type decorator such as `@minValue(...)` is accepted. The identifier between the `@` and `(`
+characters MUST refer to a user-defined function that accepts at least one argument corresponding to the targeted type
+and returns a value assignable to the following type:
+
+```bicep
+@discriminator('kind')
+type validatorReturn =
+  | { kind: 'success' }
+  | { kind: 'failure', errorMessage: string }
+```
+
+At deploy time, the ARM engine will invoke the user-defined function with the value to be validated along with any
+specified additional arguments and inspect the function's return value to determine whether the validation passed.
+
+## Terms and definitions
+
+* **Deployment boundaries** - junctures within a deployment or template that have different expression evaluation
+environments on each side. In ARM, deployment boundaries exist surrounding inner-scoped nested deployments and
+surrounding user-defined functions.
+
+## Motivation
+
+Bicep/ARM type constraints are powerful mechanisms that enforce expectations about what shape data will take at
+structural boundaries within a deployment. However, they are limited in what they can enforce to a small number of
+predefined constraint types. Allowing user-defined functions to be invoked would permit users to apply whatever
+checks they need (so long as they can be expressed via ARM's built-in functions) before allowing a value to cross a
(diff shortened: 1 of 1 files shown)


active/0000-custom-validation-decorators.md
```diff
@@ -0,0 +1,242 @@
...
+        "namespace": "_1",
+        "name": "anImportedValidator"
+      }
+    }
+  },
+  "resources": {}
+}
+```
+
+## Drawbacks
```
levimatheri, line 239: I suppose another drawback is that the decorator isn't "evaluated" until deploy time, unlike existing decorators; meaning users don't get validation at compile time for values that are known.

active/0000-custom-validation-decorators.md
```diff
@@ -0,0 +1,242 @@
...
+
```
levimatheri, line 240: just wondering, would a `throw()` function potentially help with this? When evaluated, the function would throw a `TemplateValidationException` in the backend.

active/0000-custom-validation-decorators.md
jeskew: This seems like an important limitation. It would be possible evaluate the decorator if a literal value is passed in, but we wouldn't be able to do the same kind of type-based evaluation that we do with `minValue` and the like. This will also be true of at least one other planned decorator, though (`pattern` for checking regex matches)

active/0000-custom-validation-decorators.md
```diff
@@ -0,0 +1,242 @@
...
+Users will need to declare the `validatorReturn` type or return an untyped object whenever they declare a custom
```
levimatheri, line 241: just wondering, would a `throw()` function potentially help with this? When evaluated, the function would throw a `TemplateValidationException` in the backend.

jeskew: I'll add an `Alternatives` section covering this. There would be two main differences in behavior:
* Constraint violations identify (via JSON path and coordinates) both the invalid data and the constraint that was violated. ARM's `fail` function identifies where it was invoked, which is a less specific version of the latter if code is reused.
* Data can violate multiple constraints, but `fail` function invocations can't be accumulated

levimatheri: Makes sense, thanks for clarifying

active/0000-custom-validation-decorators.md
jeskew, line 1: Feedback from team discussion:
1. Use a simpler return type to distinguish between success and failure, such as `string?` (where a `string` is an error message, and `null` means that the validation passed). We can always switch to an object return type for richer metadata if the need arises without breaking backwards compatibility.
2. Instead of inferring which functions should be usable as decorators, require an explicit opt-in. A `@validator()` decorator would work.
3. Validation logic can be complex and UDFs can only be a single expression -- will we need a `switch` or `match` expression to make this useful?

majastrz:
Would we be able to evaluate the user defined constraints on parameter default values in .bicep files and on values in bicepparam files?

majastrz (approved):
LGTM

jeskew:
> Would we be able to evaluate the user defined constraints on parameter default values in .bicep files and on values in bicepparam files?

If the default value or bicepparam value is a known at compile time, yes. We wouldn't be able to evaluate in cases where a non-deterministic function is used for a default value or when a late-bound value (one using external inputs) is used in a bicepparam file.

anthony-c-martin (approved):
LGTM!

active/0000-custom-validation-decorators.md
```diff
@@ -0,0 +1,244 @@
...
+
+```json
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "languageVersion": "2.0",
+  "parameters": {
+    "p": {
+      "type": "string",
+      "validate": [
```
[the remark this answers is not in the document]
jeskew: Would your preference be to have something like the following instead?

```json
{
  "parameters": {
    "p": {
      "type": "string",
      "@options": {
        "validate": [ ... ]
      }
    }
  }
}
```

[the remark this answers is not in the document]
anthony-c-martin: I don't have a strong feeling either way - I think what you have proposed also seems fine. `@options` was only introduced for resources to avoid naming conflicts with valid top-level resource properties (which may expand over time). This may not be necessary for other elements as we control the full schema.

PR #11, opened 2024-11-26
Bicep version pinning

levimatheri:
Partially addresses https://github.com/Azure/bicep/issues/12290

### DIFF
--- a/active/0000-bicep-version-pinning.md
+++ b/active/0000-bicep-version-pinning.md
@@ -48,48 +48,6 @@ The following example shows the proposed `version` property syntax to be added i
 
 The appropriate Bicep version will be determined by using the closest `bicepconfig.json` file to the Bicep file being compiled, merged with the default configuration.
 
-#### Cross-module compilation
-
-Each module can be constrained to a different version requirement specified in the closest `bicepconfig.json` to it. This will be beneficial for version ranges, allowing a module author to set a minimum version for compatibility if the module uses features introduced after said version. 
-
-Having said that, compilation of modules with conflicting version requirements should fail – if a module's Bicep version constraint does not match the installed Bicep binary, an error should be surfaced within the module and wherever the module is referenced. The following example illustrates this concept:
-
----------------
-**Given the installed version is v0.31.92:**
-
-_bicepconfig.json_
-```json5
-{
-    "bicep": {
-        "version": "0.31.92"
-    }
-}
-```
-
-_modules/bicepconfig.json_
-```json5
-{
-    "bicep": {
-        "version": "0.15.0"
-    }
-}
-```
-
-_modules/mod.bicep_
-```bicep
-resource otherResource 'Test.RP' = {} // <-- Bicep version "0.31.92" was used for compilation, but version "0.15.0" is required in configuration file (BCPxxx)
-```
-
-_main.bicep_
-```bicep
-module mod 'modules/mod.bicep' = { // <--  The referenced module has errors.bicep (BCP104)
-    name: 'mod'
-}
-```
-
-
-No specific change to the codebase is expected to achieve this, as Bicep already handles this today (i.e. if there are errors in a referenced module, an error is surfaced on all the references).
-
 ----------------
 
 ### Version constraint syntax
@@ -133,7 +91,7 @@ Examples:
 - `^1` - equivalent to `>=1.0.0 <2.0.0`
 
 
-### Update Bicep CLI 
+### Update Bicep compiler 
 #### Version check during compilation
 The Bicep CLI should be updated to check for the version constraint in the configuration.
 
@@ -142,6 +100,24 @@ If a user runs Bicep CLI directly to build a bicep file, we will try to resolve 
 - If the resolved version constraint does not match the locally-installed Bicep version, the compilation should fail with a clear message indicating the constraint violation.
 - If no version constraint is specified anywhere in the `bicepconfig.json` resolution chain, the compilation should fallback to using the currently installed version.
 
+#### Linter rules
+- Since we've chosen to use bicepconfig.json for version constraints, the current expectation is consistent resolution logic (e.g., the closest `bicepconfig.json` to a Bicep file is used). 
+- This also means users may specify different version constraints in different `bicepconfig.json` files. 
+- External tools do not understand how Bicep source files are grouped, therefore they can't detect version constraints in referenced modules. 
+- External tools can parse the version constraint for the entrypoint Bicep file, but conflicts won't be detected in referenced modules until the installed Bicep CLI is invoked.
+
+- In order to help the users prevent such runtime issues, we will two linter rules:
+1. **Incompatible Module Constraints:**
+    - If the version constraints specified in `bicepconfig.json` files across modules cannot be combined into a single valid version range (i.e. disjoint ranges), emit an error. For example:
+    - Module A requires `>=0.31.0 <0.32.0`.
+    - Module B requires `>=0.15.0 <0.16.0`.
+    - These constraints cannot be satisfied simultaneously, so the emit an error.
+
+2. **Looser Constraints in a parent module:**
+   - If the version constraint for the bicep file being edited is looser than any of the referenced modules' constraints, emit an error. For example:
+   - Entrypoint file requires `>=0.15.0`.
+   - Referenced module requires `<0.17.0`.
+   - Because the external tool only looks at the entrypoint file, it may download a version that is incompatible with a referenced modules (e.g. version `0.21.0` in the example).
 ----------------
 
 ### VSCode experience
@@ -161,13 +137,13 @@ We will need to clearly document this behavior in which the pinning is not stric
     - For example, given the version constraint `>=0.31.0 <0.32.0` and the available Bicep release tags `[0.21.14, 0.31.1, 0.31.92, 0.32.45]`, AzCLI would download version `0.31.92`.
 - If no version constraints are found, AzCLI should default to using locally installed version (or downloading the latest version if no Bicep is not installed locally).
 
-### Note about `use_binary_from_path` config value
+#### Note about `use_binary_from_path` config value
 When `use_binary_from_path` AzCLI config is set to `true`, or `if_found_in_ci`, AzCLI will invoke the bicep binary found in the PATH environment variable. With version pinning, the version in PATH would not necessarily correspond to the constraint resolved from the `bicepconfig.json`. 
 
 We should therefore set `use_binary_from_path` to `false` if a version constraint is found (this is similar to the behavior when a user runs `az bicep install --version` to install a specific version of Bicep. See more details on [this PR](https://github.com/Azure/azure-cli/pull/25541)).
 
-### Version syntax parsing
-Since there is no official standard for the syntax of semver ranges, different tools/libraries that implement the npm-like syntax often have subtle or major differences. As a result, we will implement our own parser in Bicep so that we can better control what is supported & not supported. We can then utilize Copilot to help translate the parsing logic to Python in AzCLI, and to Go in Ev2 CLI.
+#### Version syntax parsing
+Since there is no official standard for the syntax of semver ranges, different tools/libraries that implement the npm-like syntax often have subtle or major differences. As a result, we will implement our own parser in Bicep so that we can better control what is supported & not supported. We can then utilize Copilot to help translate the parsing logic to other languages, i.e. Python for AzCLI.
 
 ## Drawbacks
 - **Degraded visibility into registry module version constraints:** When inspecting published module source bicep file, it wouldn't be immediately clear what version was used to compile the module just by looking at the bicep file. This is because version information would be codified in the `bicepconfig.json` which is not published with the sources. However, the user can still discern the version used with the aid of the VSCode extension; the user can navigate to the compiled JSON which contains metadata about which version was used to compile the module. 
@@ -293,5 +269,5 @@ However, we should impose the following restrictions:
 
     
 ## Out of scope
-- Making this work in AzPwsh would be nice to have as the owning team is now more open for us to make changes, but would involve significant work to make it at par with AzCLI. This can be considered future enhancement.
-- There were concerns about how this would work with bicepconfig.json resolution mechanism. Today, the closest bicepconfig.json file is used to resolve configurations relative to the bicep/bicepparam file being compiled. Since there exists no merge process for the bicepconfig.json files, this could be problematic in the cases where users want to have repo-wide settings honored even with child folders containing their own bicepconfig.json files. This limitation, however, has existed already, and hence orthogonal to this proposal; we could adopt this proposal without necessarily solving the bicepconfig.json limitation. 
+1. Making this work in AzPwsh would be nice to have as the owning team is now more open for us to make changes, but would involve significant work to make it at par with AzCLI. This can be considered future enhancement.
+2. There were concerns about how this would work with bicepconfig.json resolution mechanism. Today, the closest bicepconfig.json file is used to resolve configurations relative to the bicep/bicepparam file being compiled. Since there exists no merge process for the bicepconfig.json files, this could be problematic in the cases where users want to have repo-wide settings honored even with child folders containing their own bicepconfig.json files. This limitation, however, has existed already, and hence orthogonal to this proposal; we could adopt this proposal without necessarily solving the bicepconfig.json limitation. 
--- a/active/0000-bicep-version-pinning.md
+++ b/active/0000-bicep-version-pinning.md
@@ -34,8 +34,7 @@ This approach poses a significant risk because when a new Bicep version is relea
 Implementing Bicep version pinning would ensure reproducible Bicep compilations across all deployment workflows, whether executed locally or within CI/CD pipelines. This feature would enhance stability and predictability, providing users with greater control over their deployment environments. Additionally, it would be crucial for enabling the use of Bicep natively in Ev2 to ensure safe deployment practices.
 
 ## Detailed design
-### Client side changes
-#### 1a. Add new property in bicepconfig.json
+### Add new version property in bicepconfig.json
 
 The following example shows the proposed `version` property syntax to be added in the _bicepconfig.json_ file. See [version constraint syntax below](#version-constraint-syntax) for more details:
 
@@ -49,9 +48,9 @@ The following example shows the proposed `version` property syntax to be added i
 
 The appropriate Bicep version will be determined by using the closest `bicepconfig.json` file to the Bicep file being compiled, merged with the default configuration.
 
-#### 1b. Cross-module compilation
+#### Cross-module compilation
 
-Each module can be constrained to a different version requirement specified in the closest `bicepconfig.json` to it. This will be beneficial when version ranges support is added, allowing a module author to set a minimum version for compatibility if the module uses features introduced after said version. 
+Each module can be constrained to a different version requirement specified in the closest `bicepconfig.json` to it. This will be beneficial for version ranges, allowing a module author to set a minimum version for compatibility if the module uses features introduced after said version. 
 
 Having said that, compilation of modules with conflicting version requirements should fail – if a module's Bicep version constraint does not match the installed Bicep binary, an error should be surfaced within the module and wherever the module is referenced. The following example illustrates this concept:
 
@@ -64,13 +63,6 @@ _bicepconfig.json_
     "bicep": {
         "version": "0.31.92"
     }
-}
-```
-
-_main.bicep_
-```bicep
-module mod 'modules/mod.bicep' = { // <--  The referenced module has errors.bicep (BCP104)
-    name: 'mod'
 }
 ```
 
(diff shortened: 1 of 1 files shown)


active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,134 @@
...
+```
+This approach provides more granular control over the Bicep version used for each individual file, which makes it clear which version is required for a specific file, which could be useful in scenarios where different files need to be compiled with different versions. It would also be easier to discern what version was used to compile a published module by inspecting the source.
+
+However, the downside with this approach is if usage of the same version for multiple Bicep files is desired, it would require specifying the `version` in each file which can be a maintenance overhead.
+
+## Rollout plan
+
+This proposal only involves client-side changes, and will be an opt-in feature enabled when the bicep.version configuration property is specified (essentially acting as the feature flag).
+
+## Unresolved questions
```
majastrz, line 124: What happens when there's a version constraint in bicepconfig.json and the user runs Bicep CLI directly to build the file when the version doesn't match the constraint?

levimatheri: In that case, the Bicep CLI should halt the build process and throw an error indicating that the version constraint is not satisfied (ideally along with the path to the bicepconfig.json). Does that align with your expectation? 
I have this under the [Update Bicep compiler section](https://github.com/Azure/bicep-reps/blob/976520790524b45d490959bdb2dc1b500fa633b7/active/0000-bicep-version-pinning.md?plain=1#L100), but can elaborate more there.

majastrz: What worries me is users who are interacting with multiple repos that may be owned by different people and pinned to different Bicep versions. Then, a single local installation of Bicep won't work very well.

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,135 @@
...
```
anthony-c-martin, line 125: It would be worth noting down what the user experience should be in VSCode if there's a mismatch. Do we need to make the user aware of this?

active/0000-bicep-version-pinning.md
majastrz: I think we need to confirm with the community to see how much they're using Bicep CLI as compared to az CLI.

active/0000-bicep-version-pinning.md
majastrz: If we're ok with just basic warning that the configured version doesn't match in VS code, then the implementation would be super simple. Otherwise, we'd have to find a way to decouple Bicep version from the VS code extension somehow, which will likely be very expensive.

levimatheri: Added a section on this, seems basic warning is our best option here.

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,211 @@
...
+1. The syntax is easier to read and understand as the symbols used are commonly used for comparison operations within Bicep and other tools.
+2. Would require minimal or no learning curve for customers coming from an npm or terraform background
+3. Many programming languages, including .NET ([semver package](https://semver-nuget.org/v3.0.x/)), Python ([packaging library](https://packaging.pypa.io/en/stable/specifiers.html#packaging.specifiers.SpecifierSet)), Javascript ([node-semver](https://github.com/npm/node-semver)), have libraries that provide functions to parse this syntax.
+
+*Disadvantages*:
+
+1. Could be considered more verbose in certain scenarios, e.g. to specify Bicep to match any patch version of `0.31`, the syntax would be `>=0.31.0 <0.32.0`. Likewise to match any minor version of `0`, the syntax would be `>=0 <1`.
+
+
+**Recommendation** - Due to the vast usage of the "npm/terraform syntax", and how many more tools support parsing this syntax, it seems like the better choice here.
```
levimatheri, line 154: I'm open to any objections to this recommendation.

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,231 @@
...
+
+An acceptable solution here is: instead of throwing compiler errors, we'd warn the user if their VSCode extension version doesn't match the configured version constraint. 
+This way the user is aware of the mismatch, but we do not completely block them from proceeding with compilation.
+We will need to clearly document this behavior in which the pinning is not strictly enforced in VSCode, so that users do not have the wrong expectations.
+
+
+### AzCLI updates
+- The AzCLI bicep module should be enhanced to interpret the `bicepconfig.json` version constraint.
+- When a command that requires compilation is invoked (e.g. `build`, `build-params`, `lint`, `publish`, `restore`), find the appropriate `bicepconfig.json` file by recursively searching upwards from the current directory. 
+    - If a constraint is found, download the latest version that satisfies the constraint by introspecting Bicep release tags. For example, given the version constraint `>=0.31.0 <0.32.0` and the available Bicep release tags `[0.21.14, 0.31.1, 0.31.92, 0.32.45]`, AzCLI would download version `0.31.92`.
```
levimatheri, line 199: Worth caching the versions e.g in `.azure/bin/<version>/bicep.exe` and copying to `.azure/bin/bicep.exe` when needed?

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,253 @@
...
+We will need to clearly document this behavior in which the pinning is not strictly enforced in VSCode, so that users do not have the wrong expectations.
+
+
+### AzCLI updates
+#### Basic mechanism
+- The AzCLI bicep module should be enhanced to parse the appropriate `bicepconfig.json` file by recursively searching upwards from the current directory. It should then install the latest Bicep version that satisfies the specified version constraints. 
+    - For example, given the version constraint `>=0.31.0 <0.32.0` and the available Bicep release tags `[0.21.14, 0.31.1, 0.31.92, 0.32.45]`, AzCLI would download version `0.31.92`.
+- If no version constraints are found, AzCLI should default to using locally installed version (or downloading the latest version if no Bicep is not installed locally).
+
+#### Caching
```
levimatheri, line 202: It feels like we can do this with symbolic links, but they are not as easy to use in Windows.

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,253 @@
...
+- In order to minimize unnecessary downloads and optimize bandwidth usage especially for users switching between projects that have version constraints, AzCLI should cache the downloaded version on disk, for example as `~\.azure\bin\bicepVersions\bicep_<version>.exe`. Then when a locally available version is needed, switch the bicep version by copying the binary into `~\.azure\bin\bicep.exe`. The reason for copying the binary is for backward-compatibility because currently AzCLI invokes bicep commands via the binary located in `~\.azure\bin\bicep.exe`.
```
levimatheri, line 203: It feels like we can do this with symbolic links, but they seem difficult to use in Windows.

levimatheri:
Notes from Ev2 sync:
- AzCLI download caching 
  - Prefer simpler option for now (no major concerns on CDN load), and reconsider optimizing later on
- Add note about using this syntax for modules and types 
- AzPwsh - how we should handle Bicep versioning
  -  Pulls from the PATH or fails - option1
  - Probably more significant work - option2

active/0000-bicep-version-pinning.md
levimatheri: The feedback we received was most users are using Az CLI. Thinking we can prioritize our efforts there, and then revisit the Bicep CLI use-case for future?

levimatheri:
Discussion 2/19:
- We can go ahead with az cli implementation. Based on the feedback we get from users, we can consider bicep cli experience improvement. 
- version range syntax: 
   - tied between npm/terraform (option 2) and wildcard syntax (option 3)
   - "I like the simplicity of option 3 but the fact that option 2 seems like it has more flexibility/nuance worries me a bit about 3"
   - "I'm voting for option 3 purely because of the OCI argument Marcin made...I think if we can rationalize how option 2 works with OCI, I'd vote for optin 2"

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,381 @@
...
+            '~2.1.1' // max satisfying = 2.1.2
+        ]
+
+        // return return { maxVersion: '2.1.2' }
+        ```
+
+- Other ideas for commands that could be added later on:
+    - `bicep versioning min-compatible -f <file> --versions <array of versions>`
+    - `bicep versioning satisfies --version <version> --range <range>`
+------
```
levimatheri, line 217: After speaking with @anthony-c-martin, this is costly engineering-wise, and we should prefer having tools write the bicepconfig resolution and parsing logic in their respective codebases.

levimatheri: The goal is to have a simplified model for resolving bicepconfig and parsing version that other tools can easily implement.  Copilot can help with accelerating conversion between different languages.
Similar scenario was encountered for rendering what-if changes. The logic is spread across azcli, azpwsh, and bicep-deploy.

audhage: Just a comment that might be relevant in this discussion. 

https://github.com/Azure/bicep/issues/12290#issuecomment-2890713727

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,273 @@
...
+
+```json5
+{
+    "bicep": { // section for bicep compiler options
+        "version": "0.31.92" // semantic version syntax or range to match Bicep version number
+    }
+}
+```
+
+The appropriate Bicep version will be determined by using the closest `bicepconfig.json` file to the Bicep file being compiled, merged with the default configuration.
```
jtracey93, line 65: This puts the Bicep CLI version constraint in a side-car `bicepconfig.json` rather
than in the `.bicep` source itself. Terraform's equivalent
(`required_version` in a `terraform {}` block) lives inside the `.tf` file being
compiled; see
https://developer.hashicorp.com/terraform/language/block/terraform#required_version.

I see this alternative is considered and rejected in the "Alternatives" section
(specifying `bicep.version` per-file), with the rationale that it'd need repeating
in every `.bicep` file. But Terraform's actual pattern only requires
`required_version` once per root module: it merges the setting across all
top-level blocks in that module's `.tf` files, so it's not a per-file cost in
practice. Did the team consider an equivalent "once per entrypoint/module" inline
option rather than "once per file," and if so, what ruled it out beyond the
bicepconfig.json precedent already set for extensions?

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,273 @@
...
+- `~1.2` - equivalent to `>=1.2.0 <1.3.0`
+- `~1` - equivalent to `>=1.0.0 <2.0.0`
+
+#### Caret Ranges
+Uses `^` to allow minor and patch-level changes.
+
+Examples:
+- `^1.2.3` - equivalent to `>=1.2.3 <2.0.0`
+- `^1.2` - equivalent to `>=1.2.0 <2.0.0`
+- `^1` - equivalent to `>=1.0.0 <2.0.0`
```
jtracey93: Worth stress-testing this against Terraform's `~>` before locking it in, because I
think there's a real footgun here, not just a stylistic difference.

Terraform's `~>` changes behavior based on how many version components you give it:
- `~> 1.2.3` (3 components) -> patch-level flexibility only: `>=1.2.3 <1.3.0`
- `~> 1.2` (2 components) -> minor **and** patch flexibility: `>=1.2.0 <2.0.0`

This proposal's `~` (line 82) gives `~1.2` -> `>=1.2.0 <1.3.0`, patch-only, same as
the 3-component Terraform case, but for the 2-component input where Terraform
allows a full minor range.

So `~1.2` looks identical to Terraform's `~> 1.2` but resolves to a narrower range.
A user with Terraform muscle memory who types `~1.2` expecting "any 1.2.x or newer
1.x" will silently get "only 1.2.x" instead: no error, just an unexpectedly tight
constraint. Given how many of our users work in both ecosystems (AVM ships parity
Bicep/Terraform modules), I'd expect this to generate confusion or support asks.

Two ways I can see to address it:
1. Pick a visually distinct token for the "minor-or-patch-depending-on-precision"
   behavior instead of reusing `~`/`^`, so there's no false-familiarity trap, or
2. Explicitly document, prominently (not just in this REP but in user-facing docs),
   a side-by-side table of Bicep vs. npm vs. Terraform semantics for `~`/`^`/`~>`,
   since option 2 (npm syntax) was chosen partly *because* of assumed familiarity;
   that familiarity claim needs the caveat spelled out where users will see it.

Happy to help pressure-test whichever direction you lean towards.

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,273 @@
...
+*Tools that can parse this syntax*:
+    - .NET ([semver package](https://semver-nuget.org/v3.0.x/)) - this would be useful for parsing in the Bicep & Ev2 .NET SDK codebase
+    - Python ([packaging library](https://python-semanticversion.readthedocs.io/en/latest/guide.html#the-simplespec-scheme)) - this would be useful for parsing in the AzCLI codebase
+    - Javascript ([node-semver](https://github.com/npm/node-semver)) - would be useful for bicep-node
+    - Go ([semver package](https://pkg.go.dev/github.com/Masterminds/semver/v3@v3.3.1)) - this would be useful for parsing in the Ev2 CLI codebase
+
+*Advantages*:
+
+1. The syntax is easier to read and understand as the symbols used are commonly used for comparison operations within Bicep and other tools.
+2. Would require minimal or no learning curve for customers coming from an npm or terraform background
```
jtracey93: Following on from the comment on the Tilde/Caret Ranges section above: I don't
think this claim holds for the Terraform-background case specifically. `~1.2`
under this proposal and `~> 1.2` in Terraform resolve to different range widths for
the same-looking input (patch-only here vs. minor+patch in Terraform), so a
Terraform user's existing intuition would actively mislead them rather than
transfer cleanly. Could we soften this to "npm background" only, or add the caveat
inline here so it doesn't read as a validated claim?

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,273 @@
...
+Users can use `*` to indicate that anything in the specified segment of the version is acceptable. This can be used as syntactic sugar for the version comparison syntax. The following are some examples:
+- `1.*` - equivalent to `>=1.0.0,<2.0.0`
+- `1.2.*` - equivalent to `>=1.2.0,<1.3.0`
+- `*` -  equivalent to `>=0.0.0`
+
+However, we should impose the following restrictions:
+1. The wilcard `*` cannot be used together with the version comparison syntax. 
+    - Reason: This is to prevent ambiguous snytaxes like this: `>=1.*`. It is hard to reason about the allowed range in this case, i.e. does the range start from `1.0.0` or `1.1.2` or `1.3.5` or something else?
+2. If a version number is replaced with a wildcard, then all later version numbers must also be wildcards. Thus `2.*.6` is an invalid wildcard version. 
+    - Reason: Most semver parsing tools (including the [tools sampled above](#option-2-npmterraform-syntax)) either do not support this, or they ignore anything after the `*`.
```
jtracey93: Minor: this link points to `#option-2-npmterraform-syntax`, which doesn't resolve
to any heading in the current doc; the real heading (line 190) is "Option 2: npm
syntax". Looks like this section was originally "npm/terraform syntax" (per the
Feb 2025 discussion thread on this PR) and got renamed without updating the anchor.
Worth fixing as a drive-by, and maybe worth double-checking whether anything else
in the Terraform-comparison reasoning got dropped in that same rename.

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,273 @@
...
+    - If the version constraints specified in `bicepconfig.json` files across modules cannot be combined into a single valid version range (i.e. disjoint ranges), emit an error. For example:
+    - Module A requires `>=0.31.0 <0.32.0`.
+    - Module B requires `>=0.15.0 <0.16.0`.
+    - These constraints cannot be satisfied simultaneously, so the emit an error.
+
+2. **Looser Constraints in a parent module:**
+   - If the version constraint for the bicep file being edited is looser than any of the referenced modules' constraints, emit an error. For example:
+   - Entrypoint file requires `>=0.15.0`.
+   - Referenced module requires `<0.17.0`.
+   - Because the external tool only looks at the entrypoint file, it may download a version that is incompatible with a referenced modules (e.g. version `0.21.0` in the example).
```
jtracey93: Quick terminology flag: "module" here means "a referenced `.bicep` file in the
compilation graph with its own `bicepconfig.json`," but elsewhere in the Bicep
ecosystem "module" means a published, versioned registry dependency
(`module foo 'br/public:.../foo:1.2.3'`). Since this REP explicitly doesn't cover
registry module version pinning (see comment on Out of Scope below), using the
same word for both concepts in the same document is going to cause confusion;
readers skimming this section could reasonably think it's about pinning registry
module versions. Could we rename to something like "referenced Bicep file" or
"dependency file" here to keep the two concepts visually distinct?

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,273 @@
...
+
+5. Should we build our own parsers or use libraries/packages?
+
+    ✅ Start with libraries/packages as long as they are consistent with each other and support all that we need. Reconsider writing our own if there are concerns.
+
+    
+## Out of scope
+1. Making this work in AzPwsh would be nice to have as the owning team is now more open for us to make changes, but would involve significant work to make it at par with AzCLI. This can be considered future enhancement.
+2. There were concerns about how this would work with bicepconfig.json resolution mechanism. Today, the closest bicepconfig.json file is used to resolve configurations relative to the bicep/bicepparam file being compiled. Since there exists no merge process for the bicepconfig.json files, this could be problematic in the cases where users want to have repo-wide settings honored even with child folders containing their own bicepconfig.json files. This limitation, however, has existed already, and hence orthogonal to this proposal; we could adopt this proposal without necessarily solving the bicepconfig.json limitation.
```
jtracey93: This REP only addresses Bicep CLI/compiler version pinning, but that boundary is
never stated explicitly; it's only inferable from an aside in Unresolved Question
3 ("future module/extension version range addition"). Given this is going to be
the reference doc people point to when asking "does Bicep support version pinning
yet?", can we add explicit bullets here stating:

- Registry module version-**range** pinning is out of scope (today's exact-tag
  pinning via the module source reference, e.g. `br/public:.../foo:1.2.3`, is
  unaffected and unchanged by this REP).
- Extension version-**range** pinning is out of scope (today's exact-tag pinning
  via the `bicepconfig.json` `"extensions"` map is unaffected and unchanged).
- Whether/when a follow-up REP will extend this same syntax to modules and
  extensions.

This would remove the ambiguity that's currently only resolvable by reading
between the lines.

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,273 @@
...
+Start Date: 2024-11-20
+Feature Status: Private Preview
+Bicep Issue Number(s): [#12290](https://github.com/Azure/bicep/issues/12290), [#8978](https://github.com/Azure/bicep/issues/8978)
+---
+
+# Bicep version pinning
+
+## Summary
+
+This document proposes a feature that enables users to pin specific versions of Bicep for their deployments.
```
jtracey93, line 13: Given the discussion this proposal has already generated around modules and
extensions (see the Unresolved Questions and the linked PR-comment threads), I
think this REP would benefit from an explicit "Scope" subsection right up front,
before Motivation, that enumerates the distinct version-pinning surfaces in Bicep
today and marks which this REP covers:

1. **Bicep CLI/compiler version**: this REP.
2. **Registry module version**: exact-tag pinning exists today (inline in the
   module reference); range support is future work.
3. **Extension version**: exact-tag pinning exists today (via bicepconfig.json's
   `extensions` map); range support is future work.
4. **Core Az resource type version**: currently coupled to the CLI release, no
   independent pinning; potential future work to decouple from `bicep-types-az`.

We (AVM) maintain parity Bicep/Terraform modules at scale and get direct signal on
what users expect to be able to pin and where they get surprised; happy to help
build out 2 to 4 as fast-follow REPs if that's useful, but I think this doc should at
least name them as a map of the territory rather than only surfacing them
incidentally in Unresolved Questions.

active/0000-bicep-version-pinning.md
```diff
@@ -0,0 +1,273 @@
...
+    ✅  Feedback from community was AzCLI is used more widely. Decision was to go ahead and implement for AzCLI, and revisit Bicep CLI experience later.
+
+3. Which versioning syntax to go with?
+
+    - Some concerns brought up was consistency with future module/extension version range addition, since OCI artifact versions are arbitrary and not necessarily numeric or semver.
+    - Some ideas were brought up:
+        - Use `*` wildcard. However isn't versatile enough for non-numeric usage.
+        - Use Regex. This is powerful, but could potentially intimidating/barrier to entry for unfamiliar users.
+
+    ✅ Decision was to stick to `>,<,>=,<=` syntax for version range syntax for both Bicep and module/extension versioning. The caveat is that version ranges will not be supported for module/extension OCI artifacts not following semver. In those cases, users will need to use exact versions.
```
jtracey93, line 282: This mentions "future module/extension version range addition" as the reason for
picking a syntax that's consistent across surfaces, but I don't see extension
version pinning mentioned anywhere else in this doc or in the linked discussion
notes (which mention reusing this syntax for "modules and types," not extensions).
Given extensions already have a working exact-version pinning mechanism today
(`bicepconfig.json`'s `extensions` map), can we confirm explicitly here that the
same range syntax is intended to extend to extensions too, and reconcile this with
the Out-of-Scope section (see comment there)?

jtracey93 (commented):
Thanks for the thorough writeup and for working through the syntax alternatives so
transparently. The Unresolved Questions section is genuinely useful for seeing how
the team got here.

Before going through the doc line-by-line, three higher-level things I'd like to
flag, with inline comments below on each:

1. **Terraform alignment**: the proposed syntax is explicitly npm-derived, and in at
   least one case (`~` with a 2-component version) it produces the *opposite* range
   width to Terraform's `~>` for what looks like the same input. Given how much
   Bicep and Terraform usage we have side-by-side today (AVM publishes parity
   modules in both), I think this deserves a harder look than "npm-like syntax most
   people recognize."
2. **Scope clarity**: this REP is CLI-version-only, but that's never stated
   explicitly: "module" is used in the Linter Rules section to mean something
   different from "registry module," and extension version pinning isn't mentioned
   at all (even though a pinning mechanism for extensions already exists via
   `bicepconfig.json`). I think this REP would be stronger with an explicit
   Scope/Non-Goals statement enumerating CLI vs. module vs. extension vs. (future)
   type-version pinning and what's in vs. out here.
3. A couple of smaller things (a stale anchor link, a terminology nit) called out
   inline.

None of this is a request to redo the syntax decision; more a request to make the
boundaries and the Terraform comparison explicit, since I think both will get asked
about again once this ships.

active/0000-bicep-version-pinning.md
levimatheri: That makes sense. I will tighten add a scoping section and clarify this only covers (1).

active/0000-bicep-version-pinning.md
levimatheri: We didn't specifically discuss "once per entrypoint module," but more generally were against version constraints in the source Bicep files for the following reasons:
- Complexity of parsing by external tools like AzCLI to figure out which version to download & install. Such tools would need to either use an AST-based parse (which needs a Bicep binary; a chicken-and-egg problem when the tool's job is to fetch that binary), or regex-grep the raw bicep file e.g. how [tfenv does this](https://github.com/tfutils/tfenv/blob/de6ce2e809c155cbc5e2cfeb3b1bef151244e045/lib/tfenv-min-required.sh#L8), which is fairly brittle.
With `bicepconfig.json` it's just a simple JSON parse which is straightforward.
- Bicep's module model doesn't map cleanly to Terraform's. In Terraform a module is a directory, with all its .tf files merged into one unit, so a `versions.tf` in the directory naturally scopes to the whole module. Bicep has no such boundary, and additionally any .bicep file can be an entrypoint, so "once per entrypoint module" has no stable scope to attach to. The Bicep analog of Terraform's directory-level declaration is bicepconfig.json, which already applies once across the whole subtree no matter which file is used as the entrypoint.
- It couples two separate concerns: the infrastructure a file describes and the tooling needed to process it. A compiler-version constraint is environment/tooling metadata, not part of what's being deployed. Similar idea to how C# keeps the SDK/compiler version in global.json/.csproj rather than at the top of .cs files. bicepconfig.json is Bicep's tooling-config mechanism, so the constraint naturally fits there.

active/0000-bicep-version-pinning.md
levimatheri: Thanks for calling this out, I definitely see the potential for confusion here. 
Of the options you presented, I'd prefer option 2 rather than introducing yet another token that's specific to Bicep.
I think another alternative here is to drop `~` and `^` entirely for v1 of this feature and keep just the `>`, `<`, `>=`, `<=` comparators (the former are just syntactic sugar). We can ship the comparator-based syntax, which should cover all scenarios, and add the syntactic sugar later based on user feedback.

active/0000-bicep-version-pinning.md
levimatheri: Makes sense, will update

active/0000-bicep-version-pinning.md
levimatheri: Yes, the same version range syntax is intended to extend to extensions and registry modules in the future. Will add explicit note.

active/0000-bicep-version-pinning.md
jtracey93: I think whatever decision we make we have to make it last and cant change it in a v2 of bicep as this is fundamental language stuff. If we change it in a v2 of bicep, consumers will go nuts :)

active/0000-bicep-version-pinning.md
jtracey93: could we make it part of the v1/same implementation?

levimatheri:
Notes from 7/27 discussion (updated spec to account for the meeting outcomes):
- Decided to drop `~` and `^` tokens from version range and keep the `>,<,>=,<=` comparators. We should ask users for feedback when we release this and at that point reconsider introducing syntactic sugar
- Made note to keep implementation of version range for modules & extensions out of scope. But agreed to extend the same syntax for modules and extensions for future implementation. Agreed to use this format e.g. `module foo 'oci:foo/bar:(>=1.2.0, <2.0.0)'`

anthony-c-martin (approved):

shenglol (approved):

PR #13, opened 2025-01-23
Standardize on decorator code generation

anthony-c-martin:
Summary from the rep:

> Introduce a new standardized pattern for how a decorator in the Bicep language is converted into JSON, which we can use for future decorators and possibly also consider adding for existing ones.


active/0013-decorator-transformations.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+JSON conversion:
+```json
+{
+  "type": "Microsoft.DataFactory/factories",
+  "apiVersion": "2018-06-01",
+  "name": "foo",
+  "@decorators": {
+    "description": ["hello!"],
```
subha-sa, line 78: Does this have to be array block? Using [..] indicates arrays. Can we have objects so that if a decorators have more than one value can be modelled as objects?

anthony-c-martin: As explained offline, a goal for this proposal is to have a purely mechanical translation between Bicep <-> JSON. This would mean that we don't have to build any special logic to translate each new decorator.

The reason for array rather than object is that functions in Bicep & deployments today (which are used for decorators) are purely positional - there is no concept of named arguments.

So for example, we wouldn't be able to translate `{ "foo": "..", "bar": ".." }` in JSON deterministically back into Bicep - you would have to build some logic which understands "for decorator X, foo corresponds to position 0".

active/0013-decorator-transformations.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+JSON conversion:
+```json
+{
+  "type": "Microsoft.DataFactory/factories",
+  "apiVersion": "2018-06-01",
+  "name": "foo",
+  "@options": {
+    "description": ["hello!"],
```
shenglol, line 80: Could you provide an example of lambda parameters?

anthony-c-martin: Sure - just added the example from the wait-and-retry proposal

active/0013-decorator-transformations.md
```diff
@@ -0,0 +1,94 @@
...
+JSON conversion:
+```json
+{
+  "type": "Microsoft.DataFactory/factories",
+  "apiVersion": "2018-06-01",
+  "name": "foo",
+  "@options": {
+    "description": [
+      "hello!"
+    ],
```
shenglol, line 83: I wonder if it makes sense to handle description as a special case and still emit `metadata.description`? It doesn’t quite feel like an option, and I vaguely recall some teams using `metadata.description` for logging.

anthony-c-martin: Good point - this was just meant to illustrate the design, rather than something I'm actually planning on changing. I've updated to remove this example.

shenglol (approved):
:shipit:

PR #16, opened 2025-03-01
Support resource referencing inside lambda expressions

anthony-c-martin:


active/0016-lambda-references.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Summary
+
+Proposes a mechanim such that Bicep is able to allow resource referencing inside lambda syntax blocks.
+
+## Motivation
+
+Lambdas are currently limited in their capabilities, because of limitations in the Deployment Engine with calculating references. See #14106 for an example of some of the difficulties this creates.
+
+The introduction of symbolic name referencing gives us an opportunity to improve on this.
```
jeskew, line 18: We're also working on another solution for this issue. The expression reduction work done to minimize what-if short-circuiting also allows the backend to take expressions like the following:

```bicep
resource foos '...' = [for x in range(0, 10): {
  ...
}]

outputs lambdaOverFoos array = map(foos, x => x.properties.bar)
outputs lambdaReferencingFoos array = map(range(0, 10), i => foos[i].properties.bar)
```

and transform both outputs to:

```bicep
... = [
  foos[0].properties.bar
  foos[1].properties.bar
  foos[2].properties.bar
  foos[3].properties.bar
  foos[4].properties.bar
  foos[5].properties.bar
  foos[6].properties.bar
  foos[7].properties.bar
  foos[8].properties.bar
  foos[9].properties.bar
]
```

Even if the `resources()` function already existed, we would still make this change because it will allow us to use a resource collection as the array being mapped over and because it will also support `.list*` functions (although it won't be able to handle expressions like `map(foos, f => reference(f.properties.baz, '<apiVersion>'))`). But since `resources()` doesn't yet exist, do we want to support both?

anthony-c-martin: Thanks for pointing this out! Some stuff I wasn't clear on:
* Has this already been built, or are you saying this is something we could build given the short-circuiting work?
* I didn't understand the example you gave of why `map(foos, f => reference(f.properties.baz, '<apiVersion>'))` wouldn't work - could you elaborate more?
* I would assume that the unrolling logic is still best-effort and there are various limitations - e.g. iterating over a list of runtime length. Would it be clear to the user why certain scenarios are/aren't possible?

jeskew: * This has already been built, but we haven't enabled it. Essentially, we would need to use the expression reduction instead of optimistic evaluation at one place in the template preprocessing engine (a couple lines difference). Being able to do this was one of the goals of building the expression reduction engine (though the primary motivation was improving what-if and preflight).
* The expression reducer can unroll loops and lambda constructs, but the deployments engine will still balk at nested references. `map(foos, f => reference(f.properties.baz, '<apiVersion>'))` would reduce to something like 
```json
[
  "[reference(reference('foos[0]').baz, '<apiVersion>')]",
  "[reference(reference('foos[1]').baz, '<apiVersion>')]",
  ...
]
```
which the engine would consider an error because it can't infer any implicit dependencies. One advantage of the `resources()` function is that you _can_ have nested calls like `"[resources(resources('foo').bar)]"` (although I don't think we'd have any way to decompile that into valid Bicep.
* The one limitation is that you can't have nested references. The easiest way to end up in that situation is to have a `reference` expression in the mapper function while also iterating over a list of runtime length. We should be able to detect this at compile time and raise a diagnostic. Iterating over a list of runtime length would be supported; the following works today:
```bicep
resource sa 'Microsoft.Storage/storageAccounts@2023-05-01' existing = {
  name: 'acct'
}

resource tags 'Microsoft.Resources/tags@2024-11-01' = {
  name: 'default'
  properties: {
    tags: {
      foo: string(map(sa.properties.networkAcls.ipRules, x => x.value))
    }
  }
}
```

One thing to be wary of: the deployments engine uses dependency inference to figure out which `existing` resources it can skip/optimize away. That shouldn't be an issue with Bicep-generated templates since we'd inject an explicit dependency in the right place, but it would be a footgun for anyone trying to use `resources()` in a hand-rolled JSON template. (We'd need to make sure `resources()` is only used in resource bodies to avoid skipping existing resources that are only read in a template's output declarations.)

PR #18, opened 2025-03-19
Create 0018-tooling-params-api.md

anthony-c-martin:


active/0018-tooling-params-api.md
```diff
@@ -0,0 +1,223 @@
...
+1. Deployments API removes the "expression" property and sets the "value" field for resolved parameters.
+
+## Examples
+
+### Simple env var reading
+Bicep authoring:
+```bicep
+using 'main.bicep'
+
+param foo = readInput('sys.envVar', 'MY_FOO_VAR')
```
shenglol, line 162: One idea to simplify the syntax and codegen - I wonder if it helps to introduce a `bindings` object with specific functions for different binding types:

```
param foo = bindings.env('MY_FOO_VAR')
param bar = bindings.cli('my-bar-cli-arg')
param baz = bindings.custom('__SOMETHING__')
```

Generated JSON parameters file:

```jsonc
{
  "parameters": {
    "foo": {
      "expression": "[bindings.env('MY_FOO_VAR')]"
    },
    "bar": {
      "expression": "[bindings.cli('my-bar-cli-arg')]"
    },
    "baz": {
      "expression": "[bindings.custom('__SOMETHING__')]"
    }
  },
  "bindings": {
    "env": {
      "MY_FOO_VAR": {
        // "value": "To be provided by Bicep compiler"
      }
    },
    "cli": {
      "my-bar-cli-arg": {
        // "value": "To be provided by Azure CLI / Azure Pwsh"
      }
    },
    "custom": {
      "__SOMETHING__": {
         // "value": "To be provided by the custom tool"
      }
    }
  }
}
```

anthony-c-martin: I think the problem with the above suggestion is that the "type" of binding itself needs to be more specific than just "custom", and I think some form of namespacing capability (e.g. the name of the tool) is actually useful, to reduce the chance of this getting mixed up between tools.

You could do something like using property access to denote the namespace:
```bicep
param foo = bindings.sys.envVar('MY_FOO_VAR')
param bar = bindings.sys.cliArg('my-bar-cli-arg')
param baz = bindings.toolName.bindingType('__SOMETHING__')
```

However at this point, I'd argue that it's not significantly less complex, and it introduces some "magic" (e.g. a global `bindings` variable, and the ability to resolve properties that don't really exist).

Here would be the equivalent with this proposal:
```bicep
param foo = readInput('sys.envVar', 'MY_FOO_VAR')
param bar = readInput('sys.cliArg', 'my-bar-cli-arg')
param baz = readInput('toolName.bindingType', '__SOMETHING__')
```

I definitely agree the string argument feels a bit clunky, but what I like is:
* The verb ("read") makes it a bit clearer that we are requesting something from someone.
* The first argument being a string (I think) clarifies a bit that this is a convention-based contract, not a core language concept.
* This doesn't introduce any new concepts to the language.

shenglol: That makes sense. The global bindings variable approach does feel a bit magical and hacky. Although I’m not a big fan of using strings, it might be more intuitive from a user experience perspective.

Regarding the function name, do you think bind would be a better alternative?

Also, what are your thoughts on the generated JSON object structure?

```JSON
{
  "bindings": {
    "custom": {
      "__VALUE_ONE__": {
        // "value": "To be provided by the custom tool"
      },
      "__VALUE_TWO__": {
        // "value": "To be provided by the custom tool"
      }
    }
  }
}
```

I’m proposing this format because IMHO it's more precise than:

```
{
  "bindings": {
    "0": {
      "type": "custom",
      "options": "__VALUE_ONE__"
    },
    "1": {
      "type": "custom",
      "options": "__VALUE_TWO__"
    }
  }
}
```

It may also make it easier for tools to handle, as object key lookups are generally more efficient and straightforward than array iteration.

shenglol: The above suggestion doesn't work because the value may be an object, which cannot be used as a key.

PR #22, opened 2026-07-20
Add artifact aliases and redirects REP

shenglol:
Proposes a unified artifact-reference model for Bicep modules, extensions, and local data files.

The REP defines:
- typed artifact aliases and predefined `mar` and `avm` OCI aliases
- canonical `oci:` and local `local:` artifact schemes
- deterministic local artifact redirects with named captures
- backward compatibility for existing `br`, `ts`, and `moduleAliases` references
- security, diagnostics, cache behavior, and rollout expectations

Also updates the REP template to ask authors for focused tradeoffs rather than generic drawbacks.

Related to Azure/bicep#19883.

### DIFF
--- a/active/0022-artifact-aliases-and-redirects.md
+++ b/active/0022-artifact-aliases-and-redirects.md
@@ -20,7 +20,7 @@ This proposal introduces:
 - `local`, a first-class artifact scheme for resolving artifacts from the local filesystem.
 - `oci`, the canonical scheme for registry-backed artifacts.
 
-The primary scenarios are a consistent artifact-reference experience across modules and extensions and local artifact development in monorepos. This proposal preserves existing `br` references and `moduleAliases` configuration without warnings or required migration.
+The primary scenarios are a consistent artifact-reference experience across modules and extensions and local artifact development in monorepos. This proposal preserves existing `br` references, `moduleAliases` configuration, and `moduleAliasesMock` behavior without warnings or required migration.
 
 ## Terms and definitions
 
@@ -68,6 +68,26 @@ br:mcr.microsoft.com/bicep/avm/res/storage/storage-account:0.32.1
 ```
 
 This proposal continues to support this configuration and reference syntax without warnings. It builds on the existing alias experience while making artifact references consistent for modules and extensions and adding local artifact support.
+
+### Existing module alias mock experience
+
+The existing `moduleAliasesMock` configuration, introduced by [Azure/bicep#19452](https://github.com/Azure/bicep/pull/19452), redirects a `br/<alias>:` module reference to a filesystem root while preserving the reference in Bicep source. For example:
+
+```json
+{
+  "moduleAliasesMock": {
+    "br": {
+      "company": {
+        "mapToFilePath": "./modules"
+      }
+    }
+  }
+}
+```
+
+The reference `br/company:network/vnet:1.0.0` resolves to `./modules/network/vnet.bicep` relative to the configuration file. The repository path becomes the local file path, while the tag or digest does not participate in the path. This behavior supports local testing of registry modules but applies only to `br/<alias>:` module references and always follows the fixed `<repository>.bicep` layout.
+
+`artifacts.redirects` generalizes this scenario to fully qualified and aliased modules, extensions, Template Specs, and data files, with explicit target paths and optional captures. Existing `moduleAliasesMock` configurations remain supported as described under **Incremental adoption from `br` to `oci`**.
 
 For monorepos, [Azure/bicep#19883](https://github.com/Azure/bicep/issues/19883) describes how relative references to shared artifacts become brittle when files move, while publishing every change to a registry inhibits the end-to-end development workflow that a monorepo is intended to provide. Local artifact resolution gives modules, extensions, and data files stable artifact-style references without requiring a publish step. This is the local file registry scenario from the issue; the `local` scheme in this proposal resolves files directly and does not operate an OCI registry.
 
@@ -120,7 +140,7 @@ Developers can redirect a published artifact reference to an unpublished local m
 
 ### Maintain backward compatibility
 
-Existing `br` references and `moduleAliases` configuration continue to work without modification or warnings under this proposal. Projects can adopt `oci` and `artifacts.aliases` incrementally without being required to migrate existing source or configuration.
+Existing `br` references, `moduleAliases` configuration, and `moduleAliasesMock` configuration continue to work without modification or warnings under this proposal. Projects can adopt `oci`, `artifacts.aliases`, and `artifacts.redirects` incrementally without being required to migrate existing source or configuration.
 
 ### Non-goals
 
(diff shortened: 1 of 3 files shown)


active/0022-artifact-aliases-and-redirects.md
```diff
@@ -0,0 +1,1057 @@
...
+
+#### Redirect target rules
+
+A redirect target must:
+
+- Start with `./` or `../`.
+- Resolve to exactly one local artifact file.
+- Be resolved relative to the configuration file in which the redirect is declared.
+- Refer to a `.bicep` or `.json` file for a module import.
+- Refer to a `.tgz` file for an extension import.
```
anthony-c-martin: Currently I don't think we enforce a file extension here - is this necessary?

shenglol: Oh really? If that's the case, we need to keep it consistent. Let me update the spec.

active/0022-artifact-aliases-and-redirects.md
```diff
@@ -0,0 +1,1057 @@
...
+Resolving paths relative to the declaring configuration file keeps redirects portable and makes their filesystem location unambiguous. A target can intentionally use `../` to select a sibling of the configuration directory. Path normalization uses the same rules as local artifact references; redirects do not introduce a separate filesystem sandbox.
+
+Restricting targets to local files prevents redirect chains and cycles. Redirect matching is performed once; the selected file is loaded directly.
+
+#### Capture syntax and validation
+
+A redirect key can contain named captures:
+
+- `{name}` captures exactly one artifact-reference segment and does not cross `/`, `:`, `::`, or `@` delimiters.
+- `{...name}` captures one or more `/`-separated segments and does not cross `:`, `::`, or `@` delimiters. A key can contain at most one multi-segment capture.
```
shenglol, line 559: @anthony-c-martin I added support for this because Simon mentioned that it was a much-needed feature.

anthony-c-martin: Sounds good. I'm happy with the way it's specced out here!

active/0022-artifact-aliases-and-redirects.md
anthony-c-martin: Yeah - example here - the convention I've been following is to publish to `bicep-ext-<name>` rather than `bicep-ext-<name>.tgz`: https://github.com/anthony-c-martin/bicep-ext-github/blob/8cb17db3c7dd6e65539cba5444355444f57574bd/.github/workflows/build.yml#L52

active/0022-artifact-aliases-and-redirects.md
```diff
@@ -0,0 +1,1057 @@
...
+The primary scenarios are a consistent artifact-reference experience across modules and extensions and local artifact development in monorepos. This proposal preserves existing `br` references, `moduleAliases` configuration, and `moduleAliasesMock` behavior without warnings or required migration.
+
+## Terms and definitions
+
+- **Artifact**: A Bicep module, extension package, Template Spec, local data file, or future package type that Bicep can reference or load.
+- **Artifact reference**: The source text that identifies an artifact, either through a fully qualified scheme or an alias.
+- **Artifact scheme**: A prefix such as `oci:`, `ts:`, or `local:` that identifies how Bicep resolves an artifact.
+- **Artifact alias**: A logical name that maps to an artifact location and is referenced with the `::` separator.
+- **Artifact redirect**: A development-time override that maps an artifact reference or pattern to one relative local file.
+- **MAR**: Microsoft Artifact Registry, exposed through the predefined `mar` alias.
```
majastrz: tiny nit - Could we also add something like "This was formerly known as Microsoft Container Registry (MCR)."

active/0022-artifact-aliases-and-redirects.md
shenglol: Updated.

majastrz (approved):

jtracey93 (commented):
On the whole this looks good to me and I’m supportive. 

However on the AVM front I’m not sure how the redirects would help much for local development. Especially as it seems we would not provide a way to help customers checkout a specific tag, say they cloned the BRM repo with all its tags, for each module. 

E.g. say a customer clones the entire BRM repo to their local machine and then sets up a redirect. From what I read they can only ever interact with the AVM modules .bicep file that lives in ‘main’ (or whatever branch/tag they have checked out in their clone) - is this correct?

If so I’d see value in helping customers generate .bicep files for each version or be able to pin to specific versions also in redirects so that when developing the cross AVM module references aren’t different. 

Example: let’s say a someone is developing the storage account AVM module. That has a cross module reference to the utl/types module and also the private-endpoint resource module. Let’s say both of those cross module references aren’t pointing to the latest release available of them and upgrading them is out of scope/mind for the current development efforts on the storage account module. How do we ensure we don’t introduce a silent break that the developer isn’t truly aware of through redirects not pinning to the versions of cross module dependencies? Like I show here https://youtu.be/-faq9o43tS0?t=1502&is=3mx_J4ppR1mXpHU0

—-

Other question is to confirm that it will still be possible to redirect the entire AVM artifact alias to a customers own private ACR (of the oci type) as a high level “AVM/*” wildcard pattern and then they can roll this out to all developers machines to enforce that they only ever consume AVM from their companies private ACR?

shenglol:
> However on the AVM front I’m not sure how the redirects would help much for local development. Especially as it seems we would not provide a way to help customers checkout a specific tag, say they cloned the BRM repo with all its tags, for each module.

Understood. It looks like the redirects feature doesn't completely solve the AVM test scenario, since AVM isn't a typical monorepo where cross-module references always target the current version.

One workaround I can think of is to use `git show` to extract a specific tagged version of a module into a directory outside the AVM repo:

```sh
git show v1.2.0:path/to/modules/foo/main.json > /test_modules/foo/main.json
```

We can then configure a redirect that maps the module reference to that local file. This seems like something we could automate as part of the development workflow.

> Other question is to confirm that it will still be possible to redirect the entire AVM artifact alias to a customers own private ACR (of the oci type) as a high level “AVM/*” wildcard pattern and then they can roll this out to all developers machines to enforce that they only ever consume AVM from their companies private ACR?

Yes, it is still possible. Customers can override the `avm` alias to reference modules from their private ACR,

SimonWahlin:
I really like the proposal. 

Even if the redirect feature might not solve all module development challenges, I think it will be very helpful in scenarios where I have one module referencing a second module and I want to make changes to the second module while making sure the changes works as I intended when deploying something the references the first module.

To relate to the example @jtracey93 gave above (not sure if it is relevant to AVM but I'll give it a try), lets say someone is using the storageAccount module which has a dependency on types module. The developer wants to improve the utl/types or the private endpoint module for a specific scenario. They will then be able to update a local copy of the utl/types module, redirect only that module and then test the storageAccount module without having to create an acr or publish anything using the following redirect:


```json
"avm::utl/types/avm-common-types:{tag}": "./avm-common-types.bicep"
```

or even more specific if they only want to redirect a specific tag:

```json
"avm::utl/types/avm-common-types:0.7": "./avm-common-types.bicep"
```


Anything they deploy that references avm-common-types with the redirected tag(s) will then use their local file instead of what's in AVM. Of course this should be done with a bicepconfig.json in a test folder that won't risk breaking any builds of other templates.

anthony-c-martin (approved):

PR #23, opened 2026-07-29
Add proposal for Bicep configuration inheritance

shenglol:
## Summary

- Add a relative local `extends` property for `bicepconfig.json`.
- Define deterministic linear inheritance and merge precedence.
- Preserve declaring-configuration provenance for relative path values.
- Specify validation, cycle handling, caching, and editor dependency behavior.

Addresses Azure/bicep#5022.

### DIFF
--- a/active/0023-bicep-configuration-inheritance.md
+++ b/active/0023-bicep-configuration-inheritance.md
@@ -1,5 +1,5 @@
 ---
-REP Number: "0000"
+REP Number: "0023"
 Author: shenglol (Shenglong Li)
 Start Date: 2026-07-27
 Feature Status: Public
--- a/active/0000-bicep-configuration-inheritance.md
+++ b/active/0000-bicep-configuration-inheritance.md
@@ -1,0 +1,541 @@
+---
+REP Number: "0000"
+Author: shenglol (Shenglong Li)
+Start Date: 2026-07-27
+Feature Status: Public
+Bicep Issue Number(s): "[#5022](https://github.com/Azure/bicep/issues/5022)"
+---
+
+# Bicep configuration inheritance
+
+## Summary
+
+This proposal adds configuration inheritance to `bicepconfig.json` through a top-level `extends` property. A configuration can extend one local base configuration, and configurations can form a linear inheritance chain. Derived values are merged over base values using Bicep's existing configuration merge behavior.
+
+Automatic configuration discovery remains unchanged. Each Bicep or Bicep parameters file uses the nearest ancestor file named `bicepconfig.json` as its leaf configuration. That file can use `extends` to compose shared settings from files with any name.
+
+The design preserves declaring-file-relative path behavior and treats every inherited file as a configuration dependency.
+
+## Terms and definitions
+
+- **Built-in configuration**: The default configuration distributed with Bicep.
+- **Discovered configuration**: The nearest file named `bicepconfig.json` found by searching from a source file's directory through its ancestor directories.
+- **Leaf configuration**: The discovered configuration from which an inheritance chain starts.
+- **Base configuration**: A configuration referenced by another configuration's `extends` property.
+- **Configuration chain**: The ordered set containing a leaf configuration and all of its transitive base configurations.
+- **Effective configuration**: The result of merging the built-in configuration, the configuration chain, and any supported invocation-time override.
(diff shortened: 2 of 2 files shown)


active/0023-bicep-configuration-inheritance.md
```diff
@@ -0,0 +1,541 @@
...
+- **Configuration document**: One parsed configuration file together with its URI and source information.
+- **Configuration source map**: A mapping from values in the merged configuration to the configuration document and source span that declared each value.
+
+## Motivation
+
+This proposal addresses [Azure/bicep#5022](https://github.com/Azure/bicep/issues/5022), which requests TypeScript-style configuration inheritance for Bicep.
+
+Bicep currently merges one user configuration with the built-in configuration. When multiple ancestor directories contain `bicepconfig.json`, Bicep uses only the nearest file and ignores the others. This is predictable, but it requires users to duplicate shared analyzer, formatter, extension, and artifact settings whenever a nested project needs a small customization.
+
+This is especially costly in monorepos and centrally managed build systems:
```
anthony-c-martin, line 37: How often is this need coming up? It doesn't look like the linked GitHub issue has much interaction, despite being almost 5 years old

active/0023-bicep-configuration-inheritance.md
```diff
@@ -0,0 +1,541 @@
...
+
+The feature does not need an experimental configuration flag. Such a flag would itself depend on successfully loading the configuration whose loading behavior is being enabled. Use of `extends` is explicit, and configurations that omit it retain their existing behavior. Any telemetry added for the feature must not collect configuration paths or values.
+
+## Unresolved questions
+
+None at this time.
+
+## Out of scope
+
+- Multiple direct bases or diamond inheritance.
```
anthony-c-martin, line 535: How confident are we that we'll never had to change this again? With the version pinning spec, we'll need to make sure that the algorithm used here is always going to be back compatible.

shenglol: I'm fairly confident. TypeScript's tsconfig has always supported only a single base config, and that model seems to have worked well for the community. Bicep's configuration is much simpler than TypeScript's, so I expect a single-parent model will cover the vast majority of use cases.

ESLint is one of the few tools that supports multiple base configs, but I generally avoid that pattern personally because it makes configuration resolution and conflict handling much harder to reason about.

The nice thing is that starting with a single parent doesn't paint us into a corner. If we ever see compelling scenarios for multiple base configs, we can add support later in a backward-compatible way.

active/0023-bicep-configuration-inheritance.md
shenglol: Not very often honestly. That said, I'd still like to prioritize it because I think it's better to establish the inheritance model before we introduce additional `bicepconfig.json` features, such as:

- [`artifacts.redirects`](https://github.com/Azure/bicep-reps/blob/main/active/0022-artifact-aliases-and-redirects.md)
- [Version pinning](https://github.com/Azure/bicep-reps/pull/11)
- Specifying a config file when invoking Bicep CLI commands (https://github.com/Azure/bicep/issues/5013)

These features can technically be implemented independently, but having the inheritance model defined upfront gives us a clearer foundation for how configuration is resolved. Otherwise, each new feature would need to make assumptions about inheritance behavior, and we may need to revisit those decisions later to ensure they work consistently with inherited configs.

active/0023-bicep-configuration-inheritance.md
```diff
@@ -0,0 +1,541 @@
...
+    }
+  }
+}
+```
+
+`main.bicep` discovers the root `bicepconfig.json`. `modules/storage.bicep` independently discovers `modules/bicepconfig.json`, which inherits the root configuration and the shared base transitively. Existing nearest-file boundaries remain intact; inheritance between them occurs only because the nested configuration declares it.
+
+## Tradeoffs
+
+- Effective configuration becomes harder to infer from one file. Editor navigation and diagnostics that include the inheritance chain mitigate this cost.
```
levimatheri, line 494: Not sure if "Editor navigation" includes this but it would be nice to have a VSCode command to view effective config.

shenglol: That's definitely something we can implement.

harshpatel17 (approved):

levimatheri (approved):
:shipit:

anthony-c-martin (approved):

PR #26, opened 2026-08-25
Add OCI template link spec for addressing deployments 4MB limit

levimatheri:
List of Bicep issues related to this spec:
- https://github.com/Azure/bicep/issues/5890
- https://github.com/Azure/bicep/issues/12919
- https://github.com/Azure/bicep/issues/8586
- https://github.com/Azure/bicep/issues/6970
- https://github.com/Azure/bicep/issues/4293


active/0000-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
+---
+REP Number: "0000"
```
majastrz: nit - do you need to update the rep number to the next unused one?

active/0026-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
...
+1. The user then invokes tooling to submit a deployment request with the OCI reference as the `templateLink`.
+1. The Deployments service downloads the OCI artifact, validates its manifest and layers, and deploys it.
+
+#### Consuming
+
+##### Authoring experience
+
+Referencing an OCI module in a Bicep file:
+
+1. The user references a multi-layered registry module using the existing `br:` syntax.
```
majastrz, line 67: Would we allow cross-registry references?

active/0026-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
...
+bicep deploy --resource-group <rg> --template-oci-reference <registry>/<repo>:<tag>
+az deployment group create --resource-group <rg> --template-oci-reference br:<registry>/<repo>:<tag>
+New-AzResourceGroupDeployment -ResourceGroupName <rg> -TemplateOciReference br:<registry>/<repo>:<tag>
+```
+
+`bicep deploy` and `az deployment group create`/`New-AzResourceGroupDeployment` gain a `--template-oci-reference` parameter (mutually exclusive with `--template-file`/`--template-uri`/`--template-spec`). Publishing and deploying remain two distinct, explicit steps — neither command publishes on the caller's behalf. See [Artifact format](#artifact-format) for what `--artifact-format layered` actually produces.
+
+### Artifact format
+
+`bicep publish --artifact-format layered` publishes one manifest whose `config.mediaType` is `application/vnd.ms.bicep.module.config.v2+json`. It uses the same `artifactType` as the existing single-file format because both are Bicep module artifacts with different internal layouts:
```
majastrz, line 114: What's the value of `--artifact-format` for the current single layer format?

active/0026-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
...
+      "location": "eastus",
+      "properties": { "addressSpace": { "addressPrefixes": ["10.0.0.0/16"] } }
+    }
+  ]
+}
+```
+
+### Authentication and registry support
+
+**CI/CD considerations:**
```
majastrz, line 206: Should we call out that if someone has CI/CD setup for container images, there should be no permissions changes needed? (Assuming that's true.)

active/0026-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
...
+
+##### Option 2: OCI image index
+
+Publish an [OCI image index](https://oras.land/docs/concepts/artifact/#harnessing-image-indexes) referencing both the existing single-file manifest and the new layered manifest under one tag.
+
+- Pros:
+  - Once index support and a Bicep-specific manifest-selection algorithm ship, one tag can expose both formats to index-aware clients.
+  - Matches how the [OCI spec distinguishes a manifest from an index](https://github.com/opencontainers/image-spec/blob/main/manifest.md): each manifest still describes one specific representation, and the index is the sanctioned mechanism for exposing variants from one reference.
+- Cons:
+  - Requires implementing image-index resolution and manifest selection, **which doesn't exist in Bicep today**, therefore this does not help clients released before that support ships.
```
majastrz, line 265: The `platform` property in the index seems super tied to os and architecture as per https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/image-index.md#image-index-property-descriptions, so we'd probably need to use something else to tell Bicep which one is which unless we introduce a new version for the artifact type.

active/0026-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
...
+  - Uses additional registry storage because each artifact retains both the flattened legacy template and the layered representation.
+
+This also changes the [server-side resolution logic](#server-side-changes): the manifest would need to keep the legacy `config.mediaType` of v1 for the new layered representation too, with an added `entryPointDigest` field that the server reads to find the entry-point layer. This works because existing Bicep versions already ignore the contents of the config blob.
+
+##### Option 4: Publish the layered format under a separate tag
+
+The customer publishes the layered format under a second, separate tag (e.g. `v1.2.0-layered`), in addition to the existing single-file tag they already publish. This is a manual step on the customer's side — Bicep does not publish both tags from a single `bicep publish` invocation.
+
+- Pros:
+  - Works today with no new Bicep engineering; existing clients and tags are unaffected.
```
majastrz, line 288: This seems the same as option 1. There's also a pro of this that puts the user in control of what happens both when publishing and when consuming. It's also very easy to understand without any behind the scenes "magic".

active/0026-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
...
+- **Compile** builds the module's semantic model from the entry-point layer, whose parameters, outputs, and exports are already declared there — no flattening needed for type-checking.
+- **Build** flattens the layers depending on the output format: `bicep build` (or `bicep publish` without `--artifact-format layered`) recursively inlines each nested `templateLink.digest` reference into a single template, reproducing today's flattened output; `bicep publish --artifact-format layered` instead copies the referenced artifact's transitive layers byte-for-byte into the new manifest, deduplicated by digest.
+
+An artifact using the existing single-file format (`config.mediaType` of `application/vnd.ms.bicep.module.config.v1+json`) is unaffected — Bicep resolves it exactly as it does today.
+
+#### Compatibility with older Bicep CLI versions
+
+A Bicep CLI version released before layered-artifact support cannot restore or consume an artifact published with `--artifact-format layered`. Four options are under consideration:
+
+| Option | Old clients can consume the artifact | Extra registry storage | Additional Bicep engineering beyond the core proposal |
```
majastrz, line 238: No matter which option we choose, the user should be in control of doubling their ACR storage.

active/0026-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
...
+### Microsoft.Resources/deployments API changes
+
+New `ociReference` and `digest` properties are added to the `templateLink` object on a deployment resource, **gated by a server-side feature flag**. At the root, `digest` is valid only when paired with `ociReference`; within a template already resolved from an OCI artifact, a digest-only `templateLink` identifies a layer in that same artifact.
+
+```json
+// ✅ ociReference alone — resolves the entry point via the artifact's manifest
+{ "templateLink": { "ociReference": "myregistry.azurecr.io/bicep/deployments/my-app:v1.2.0" } }
+
+// ✅ ociReference + digest — pins to one specific layer, skipping manifest resolution
+{ "templateLink": { "ociReference": "myregistry.azurecr.io/bicep/deployments/my-app:v1.2.0", "digest": "sha256:bbb222..." } }
```
majastrz, line 322: Is this how we will address specific layers in nested deployments during execution?

active/0026-oci-template-links.md
levimatheri: This has the potential to create some authN challenges and would mean that artifacts are not necessarily self-contained, the consensus from an earlier discussion was to block this for now but leave it open as an option for the future.

majastrz: 100% Agree. We should probably call out the restriction explicitly, though. (Unless it's already there and I missed it.)

active/0026-oci-template-links.md
levimatheri: Yes, that's right.

active/0026-oci-template-links.md
levimatheri: We can use `--artifact-format single` for the existing format and default to it if the argument is not specified.

active/0026-oci-template-links.md
levimatheri: Yes, there should be no additional permission config changes in this case, will add that.

active/0026-oci-template-links.md
levimatheri: Agreed. Another alternative I would be partial to is to use `annotations` to disambiguate.

active/0026-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
...
+**Authentication expectations:**
+
+- Publishing uses the existing `bicep publish` authentication model for module registries, such as the caller's `az login` context.
+- Consuming the artifact at deploy time does not require credentials for anonymously readable repositories. For private repositories, the Deployments service exchanges the caller's Entra ID token for an ACR token carrying only the caller's pull permissions; no separate secret or SAS management is needed.
+- This token-exchange design is intended to support both user and service-principal callers. Service-principal delegation depends on ACR implementing stronger claim validation and is required before general availability.
+
+**Registry support:**
+
+- Initially, the Deployments service resolves `ociReference` only from Azure Container Registry. The OBO-based authentication described above is ACR-specific (`AcrPull`), but the artifact format is standard OCI and contains nothing ACR-specific.
+- We will initially support ACR registries which are network-accessible to the Deployments service.
```
jaredfholgate, line 220: Does this mean that the ACR needs to have public internet access? Or could it work with a network restriction like allow trusted services (https://learn.microsoft.com/en-us/azure/container-registry/allow-access-trusted-services).

I can imagine many customers would push back on the need for the ACR with potentially sensitive data needing to be public.

active/0026-oci-template-links.md
```diff
@@ -0,0 +1,380 @@
...
+
+## Drawbacks
+
+- Requires the user to have (or create) a container registry as a prerequisite for large deployments, which is a new dependency for anyone who previously only relied on inline templates.
+- Implementing OCI authentication for Deployments-to-ACR calls requires additional engineering effort, though a viable approach (exchanging the caller's Entra ID token for a pull-scoped ACR token) is already identified.
+- Layered artifacts are incompatible with Bicep CLI versions that predate layered-manifest support; consumers must upgrade before referencing a layered artifact, unless we adopt one of the mitigations in [Compatibility with older Bicep CLI versions](#compatibility-with-older-bicep-cli-versions).
+
+## Alternatives
+
+1. **Raise the ARM request-body limit.** The 4 MB limit is a longstanding dependency of ARM PUT request processing and is unlikely to change. Raising it would also only move the boundary: sufficiently large deployments would eventually require another increase, without gaining self-containment or digest-based integrity.
```
jaredfholgate, line 340: Couldn't chunking help here? The limit could remain at 4MB, but Bicep chunks the PUT into 4MB blocks and ARM re-constitutes on the other end?

jaredfholgate (commented):
I understand why this approach is being put forward, but it feels a bit clunky from a customer perspective. They need to deploy, secure, and maintain a container registry that they don't really need. The container registry just contains ephemeral artifacts that are not really useful outside of this use case (as a far as I know), so they also need to implement a clean-up mechanism. The container registry also has to be exposed to the public internet.

Have we really exhausted all options to make this seamless from a customer perspective? Is there more we could do on the ARM side to raise this 4MB limit, such as accepting chunked http requests or similar? I appreciate there would need to be some top end limit, but wouldn't that also apply to ACR hosted images too or would we allow infinite size of ACR hosted modules to be processed by ARM? I also appreciate the push vs pull nature of this, but the same size of module ends up cached on the ARM side and I think the push would reduce the overall total network requests slightly?

I have no doubt you've already considered this before proposing this solution, so apologies if this comes across as overly critical. My perspective is as a module author and consumer.

slavizh:
I am also not fond of this approach. It replaces one limit with another limit. Consuming modules cross-tenant would be difficult or limited due to deployment being in one tenant and the container registry located in another tenant. We currently have one Container registry that is used by different tenants. Also in order to take advantage of this you will have to re-publish all your existing modules into a new format. That is not good approach. For example, if you have some modules published now, they are published with specific Bicep CLI version that works at the time the module was published. So if you want to take advantage of this you will have to re-publish all your existing modules and the version that you want via new bicep cli version.

active/0026-oci-template-links.md
levimatheri: Using trusted services for new scenarios may not be possible, the current recommendation is to use NSP (Network Security Perimeter) rules for service-to-service network access. However, ACR hasn't onboarded to NSP yet. 
We'll look into feasibility of NSP with ACR team.

